PR #24, opened 2016-06-05
added test cases which validate spec compliance

sokra:
The current implementation doesn't comply the harmony spec in important aspects:

1. imported values are live bindings
2. imported modules execute before the importing module
3. exports are bound before the module executes

> import and export statements.

You always talk about *statements*, but they ain't statements. They do their job before the normal execution of the module starts (see 2. and 3.). Just like FunctionDeclarations... You may solve this by hoisting the effects of import and exports.

> Calling these setter functions one or more times is the key to implementing live bindings, as required by the ECMAScript 2015 specification.

I think setter functions are the opposite of the solution for live bindings. In my opinion the concept of setter functions as used here prevent you from correctly implementing live bindings. I recommend to drop this concept and redesign with live bindings in mind.

### DIFF
--- a/test/live.js
+++ b/test/live.js
@@ -1,0 +1,4 @@
+export var value = 0;
+export function add(x) {
+  value += x;
+};
--- a/test/order-c.js
+++ b/test/order-c.js
@@ -1,0 +1,4 @@
+import { log } from "./order-tracker";
+import "./order-a";
+log("c");
+import "./order-b";
--- a/test/export-cycle-a.js
+++ b/test/export-cycle-a.js
@@ -1,0 +1,9 @@
+function fun() {
+  return true;
+};
+
+import { callFun } from "./export-cycle-b";
+
+export default callFun();
+
+export { fun };
--- a/test/export-cycle-b.js
+++ b/test/export-cycle-b.js
@@ -1,0 +1,5 @@
+import { fun } from "./export-cycle-a";
+
+export function callFun() {
+  return fun();
+}
--- a/test/order-a.js
+++ b/test/order-a.js
@@ -1,0 +1,2 @@
+import { log } from "./order-tracker";
+log("a");
--- a/test/order-b.js
+++ b/test/order-b.js
@@ -1,0 +1,2 @@
+import { log } from "./order-tracker";
+log("b");
--- a/test/order-tracker.js
+++ b/test/order-tracker.js
@@ -1,0 +1,7 @@
+var data = [];
+export function log(x) {
+  data.push(x);
+};
+export function getLog() {
+  return data;
+};
--- a/test/tests.js
+++ b/test/tests.js
@@ -229,6 +229,26 @@ describe("export statements", function () {
   });
 });
 
+describe("spec compliance", function () {
+  it("should establish live binding of values", function () {
+    import { value, add } from "./live";
+    assert.equal(value, 0);
+    add(2);
+    assert.equal(value, 2);
+  });
+
+  it("should execute modules in the correct order", function () {
+    import { getLog } from "./order-tracker";
+    import "./order-c";
+    assert.deepEqual(getLog(), ["a", "b", "c"]);
+  });
+
+  it("should bind exports before the module executes", function () {
+    import value from "./export-cycle-a";
+    assert.equal(value, true);
+  });
+});
+
 describe("built-in modules", function () {
   it("should fire setters if already loaded", function () {
     // The "module" module is required in ../lib/node.js before we begin


benjamn:
Thanks for your feedback, @sokra!

> I think setter functions are the opposite of the solution for live bindings. In my opinion the concept of setter functions as used here prevent you from correctly implementing live bindings.

We must have a different understanding of the semantics of live bindings, or perhaps I need to explain differently how the setter functions work, because I'm not quite following this objection. In your understanding, in what way is a live binding different from a local variable whose value can change when the exporting module defines (or redefines) it?

I take your points about hoisting, and will consider how that might work. Are there specific problems you think would not be solved by hoisting `module.import` and `module.export` calls? Obviously the whole idea of nesting imports implies that some modules may not always be imported, if the code that contains the `import` declaration is never executed, so your 2nd requirement will have to be relaxed/generalized in order for nested `import` declarations to have any value.

benjamn:
Also, just to keep the limitations of transpilers in view (because `reify` is definitely a transpiler, not a native implementation), how would you compare the spec compliance of something like Babel's CommonJS transform to that of `reify`?

test/live.js
```diff
@@ -0,0 +1,4 @@
+export var value = 0;
+export function add(x) {
+  value += x;
```
benjamn: This is a known problem that I plan to solve by calling `module.export("value")` immediately after the assignment/update. For comparison, Babel would achieve this by replacing this assignment statement with `exports.value = value += x`.

test/order-c.js
```diff
@@ -0,0 +1,4 @@
+import { log } from "./order-tracker";
+import "./order-a";
+log("c");
+import "./order-b";
```
benjamn: I see, you're expecting this `import "./order-b"` declaration to be hoisted above the `log("c")` call. I think that's a problem that would be solved by hoisting `import` declarations, or simply restricting `import` declarations so that they must appear before other statements/declarations, so this ordering would be forbidden by the compiler.

sokra:
> In your understanding, in what way is a live binding different from a local variable whose value can change when the exporting module defines (or redefines) it?

In no way. That's what it is. The way how you export it looks fine. The problem is the way you import values from other modules: You copy them once (ok, multiple times, but you don't "link" them).

Take a look at the example from http://www.2ality.com/2015/07/es6-module-exports.html

image

The transpiled code you emit only copy the value of `mutableValue` after the `import`, but accessing the `mutableValue` in `main1.js` doesn't access the variable in `lib.js`. It only reads the copy. You have two vars named `mutableValue`, one in `lib` and one in `main1`.

The way i. e. babel or webpack transpile it does access the orginal variable directly. They don't have an additional variable in `main1`.

> Are there specific problems you think would not be solved by hoisting module.import and module.export calls?

This will solve 2. and 3.

> Obviously the whole idea of nesting imports implies that some modules may not always be imported, if the code that contains the import declaration is never executed, so your 2nd requirement will have to be relaxed/generalized in order for nested import declarations to have any value.

To be consistent I would always hoist import and export inside the current scope. (This sentence doesn't imply that I take any side regarding nested imports)

> Also, just to keep the limitations of transpilers in view (because reify is definitely a transpiler, not a native implementation), how would you compare the spec compliance of something like Babel's CommonJS transform to that of reify?

As far as I know babel comply to the spec at least in this aspects. Transpilation is not a limitation here as I already implemented ES6 Modules in this way as transpilation for webpack.

test/live.js
sokra: :+1: Sounds like this would solve the live binding issue.

sokra:
> I think setter functions are the opposite of the solution for live bindings. In my opinion the concept of setter functions as used here prevent you from correctly implementing live bindings. I recommend to drop this concept and redesign with live bindings in mind.

> I plan to solve by calling module.export("value") immediately after the assignment/update.

Considering this plan, I must admin that this way the setters concept is probably possible and a very elegant way of solving the debugging problem.

sokra:
> Considering this plan, I must admin that this way the setters concept is probably possible and a very elegant way of solving the debugging problem.

After sleeping...

How do you plan to solve this cases?

* non-ES6 modules

``` js
exports.a = 1;
// ...
exports.a = 2;
```

* already transpiled ES6 modules

``` js
// i. e. by babel.js
var a = exports.a = 1;
// ...
exports.a = a = 2;
```

* undetectable writes to variables

``` js
export var a = 1;
// ...
eval("a = 2");
```

benjamn:
Thanks again for keeping me honest, and pushing me to think about these cases.

If the module being imported is a traditional CommonJS module, I think all that matters is the `module.exports` object, and live binding for local variables probably will not be supported. That said, we know what variables have been imported by other modules, so we could register setter functions for those properties on the `module.exports` object. I've completely avoided using `Object.defineProperty` in the `reify` implementation so far, because it usually leads to performance problems, but it might be worthwhile in this case.

The good thing about `eval` is that it has to be "direct" in order to manipulate local variables, so it's possible to identify risky `eval(...)` calls using a simple static analysis. Then it's just a matter of rewriting `eval(...)` to `eval(...);module.export()` to trigger another round of setters, if any of the exported values have changed.

I'll file some issues for the features/fixes you've pointed out.

trusktr:
> so this ordering would be forbidden by the compiler.

Some people might like to write code like this:

```js
import foo from 'foo'
foo()

import bar from 'bar'
bar()
```

In larger files, the co-location of imports to where the symbols are used can help keep thoughts grouped together. Such a set a grouped codes might be subject to an eventual file splitting, but at least the ideas can be grouped while the author is writing and before having to think about manipulating files.

benjamn:
@trusktr yep! that's a good point

Issue #55, opened 2016-12-08
How to get more helpful error messages


sebakerckhof:
Currently using reify, compilation can throw errors like:
```
"SyntaxError: Unexpected token (44:6)\n    at Parser.pp$4.raise (/home/seke/dev/purple/node_modules/acorn/dist/acorn.js:2221:15)
  at Parser.pp.unexpected (/home/seke/dev/purple/node_modules/acorn/dist/acorn.js:603:10)    
  at Parser.pp$2.parseBindingAtom (/home/seke/dev/purple/node_modules/acorn/dist/acorn.js:1417:12)
  at Parser.pp$1.parseVarId (/home/seke/dev/purple/node_modules/acorn/dist/acorn.js:1049:20)   
  at Parser.pp$1.parseVar (/home/seke/dev/purple/node_modules/acorn/dist/acorn.js:1032:14)    
  at Parser.pp$1.parseVarStatement (/home/seke/dev/purple/node_modules/acorn/dist/acorn.js:917:10)
  at Parser.pp$1.parseStatement (/home/seke/dev/purple/node_modules/acorn/dist/acorn.js:706:19)    
  at Parser.parseStatement (/home/seke/dev/purple/node_modules/acorn-es7-plugin/acorn-v3.js:105:25)
  at Parser.pp$1.parseBlock (/home/seke/dev/purple/node_modules/acorn/dist/acorn.js:981:25)
  at Parser.pp$1.parseStatement (/home/seke/dev/purple/node_modules/acorn/dist/acorn.js:709:33)
```

Is there any way to make this more clear? At the very least an indication in which file the error occurs would help a lot...

PR #57, opened 2017-01-17
Add filename on error

sebakerckhof:
Try to solve: https://github.com/benjamn/reify/issues/55

### DIFF
--- a/node/caching-compiler.js
+++ b/node/caching-compiler.js
@@ -23,7 +23,7 @@ exports.compile = function (content, filename) {
     return info ? readWithCache(info, content) : content;
   } catch(e) {
     if (filename) {
-      e.message += `\nWhile processing file: ${filename}`;
+      e.message += 'While processing file: ' + filename;
     }
     throw e;
   }
--- a/node/caching-compiler.js
+++ b/node/caching-compiler.js
@@ -17,8 +17,17 @@ exports.compile = function (content, filename) {
     // Treat the REPL as if there was no filename.
     filename = null;
   }
-  var info = filename ? getPkgInfo(filename) : fallbackPkgInfo;
-  return info ? readWithCache(info, content) : content;
+  
+  try {
+    var info = filename ? getPkgInfo(filename) : fallbackPkgInfo;
+    return info ? readWithCache(info, content) : content;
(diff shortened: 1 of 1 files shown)


sebakerckhof:
@benjamn :question:

node/caching-compiler.js
```diff
@@ -17,8 +17,17 @@ exports.compile = function (content, filename) {
...
+    var info = filename ? getPkgInfo(filename) : fallbackPkgInfo;
+    return info ? readWithCache(info, content) : content;
+  } catch(e) {
+    if (filename) {
+      e.message += 'While processing file: ' + filename;
+    }
+    throw e;
+  }
+
 };
```
benjamn: Performance of this function is important, so perhaps we should split it into two different helper functions, with only one function containing a `try`-`catch` block (i.e., the function called when `filename` is provided).

node/caching-compiler.js
```diff
@@ -17,8 +17,17 @@ exports.compile = function (content, filename) {
...
   }
-  var info = filename ? getPkgInfo(filename) : fallbackPkgInfo;
-  return info ? readWithCache(info, content) : content;
+  
+  try {
+    var info = filename ? getPkgInfo(filename) : fallbackPkgInfo;
+    return info ? readWithCache(info, content) : content;
+  } catch(e) {
+    if (filename) {
+      e.message += 'While processing file: ' + filename;
```
benjamn: No need for a space before `While`?

node/caching-compiler.js
sebakerckhof: You probably know way more about javascript optimization, but I thought since https://github.com/v8/v8/commit/9aac80f try-catch didn't have much of an impact anymore. And even when it did, it only concerned code at the same stack-frame, which is almost nothing here.

node/caching-compiler.js
sebakerckhof: Yeah, I first had it at a newline.

node/caching-compiler.js
benjamn: Meteor's still on V8 4.5.103.43, but I agree this might be a premature optimization for one JS engine.

PR #62, opened 2017-02-13
`Declaration`s should become `Expression`s in exported value assignment.

abernix:
This is submitted for consideration only as I'm not sure I'm qualified to speak to the far-reaching effects of this, but I can attest that problems are observed without this change as shown in the attached [Meteor issue](https://github.com/meteor/meteor#8312) and am interested in input.

Commit message included here as my best explanation, but see also [my comment on the Meteor issue](https://github.com/meteor/meteor/issues/8312#issuecomment-279367119):

----

Reify transforms default exports of `ClassDeclaration`s and
`FunctionDeclaration`s into what are essentially `ClassExpression`s and
`FunctionExpression`s.

When it comes time to use this transformed code, `acorn` appears to be
flexible when it receives a `Declaration` on the right-hand side of an
`Assignment`, but `babel` is not (it insists an `Expression`)

This change transforms the AST (when permitted by `ast: true`), to match
the assignment which is presented in the code and introduces a test to
ensure that this transformation is made.

Fixes meteor/meteor#8312

Possibly related? https://github.com/babel/babylon/issues/325

### DIFF
--- a/test/tests.js
+++ b/test/tests.js
@@ -523,24 +523,45 @@ describe("compiler", function () {
 
   it("should transform default export declaration to expression", function () {
     import { compile } from "../lib/compiler.js";
+    
+    function parse(code) {
+      var result = compile(code, { ast: true });
+      var ast = result.ast;
 
-    var code = [
+      if (ast.type === "File") {
+        ast = ast.program;
+      }
+
+      assert.strictEqual(ast.type, "Program");
+
+      return ast;
+    }
+
+    var anonCode = [
+      "export default class {}"
+    ].join("\n");
+
+    var anonAST = parse(anonCode);
+
+    assert.strictEqual(anonAST.body.length, 1);
+
+    assert.strictEqual(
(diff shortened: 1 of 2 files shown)


test/tests.js
```diff
@@ -521,6 +521,28 @@ describe("compiler", function () {
     isCallExprStmt(ast.body[5], "module", "export");
   });
 
+  it("should transform default export declaration to expression", function () {
+    import { compile } from "../lib/compiler.js";
+
+    var code = [
+      "export default class {}"
```
benjamn: Maybe give this class a name so that it's more ambiguous whether it's a declaration or an expression?

abernix: Hmm, it's specifically the lack of a name which causes the undesired result.  Named declarations are hoisted in a different manner, [in this code](https://github.com/benjamn/reify/blob/master/lib/compiler.js#L371-L379).

benjamn:
Wow, I'm not doing so well here. Sorry for the noise!

PR #75, opened 2017-03-18
Help out bundlers.

jdalton:
This PR helps bundlers out by cherry-picking and avoiding explicit id reference for conditional loads.

#### webpack.config.js
```js
'use strict'
const webpack = require('webpack')

module.exports = {
  'target': 'node',
  'plugins': [
    new webpack.optimize.UglifyJsPlugin,
    new webpack.DefinePlugin({
      'process.env.REIFY_PARSER': '"top-level"'
    })
  ]
}
```

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -1,9 +1,13 @@
 var hasOwn = Object.prototype.hasOwnProperty;
 var assert = require("assert");
-var types = require("ast-types");
+var types = require("ast-types/fork")([
+  require("ast-types/def/esprima"),
+  require("ast-types/def/babel6")
+]);
+
 var n = types.namedTypes;
 var b = types.builders;
-var MagicString = require("magic-string");
+var MagicString = require("magic-string/dist/magic-string.cjs.js");
 var utils = require("./utils.js");
 var shebang = /^#!.*/;
 
--- a/lib/parsers/default.js
+++ b/lib/parsers/default.js
@@ -1,16 +1,16 @@
+var dynRequire = module.require ? module.require.bind(module) : __non_webpack_require__;
+
 var REIFY_PARSER =
   typeof process === "object" &&
   typeof process.env === "object" &&
   process.env.REIFY_PARSER;
 
-if (REIFY_PARSER === "acorn") {
-  exports.parse = require("./acorn.js").parse;
-} else if (REIFY_PARSER === "babylon") {
-  exports.parse = require("./babylon.js").parse;
-} else if (REIFY_PARSER === "top-level") {
-  exports.parse = require("./top-level.js").parse;
+if (REIFY_PARSER === "acorn" ||
+    REIFY_PARSER === "babylon" ||
+    REIFY_PARSER === "top-level") {
+  exports.parse = require("./" + process.env.REIFY_PARSER + ".js").parse;
 } else try {
-  exports.parse = require(REIFY_PARSER);
+  exports.parse = dynRequire(REIFY_PARSER);
 } catch (e) {
-  exports.parse = require("./acorn.js").parse;
+  exports.parse = dynRequire("./acorn.js").parse;
 }
--- a/lib/runtime.js
+++ b/lib/runtime.js
@@ -1,5 +1,6 @@
 var Entry = require("./entry.js").Entry;
 var utils = require("./utils.js");
+var dynRequire = module.require ? module.require.bind(module) : __non_webpack_require__;
 
 exports.enable = function (Module) {
   var Mp = Module.prototype;
@@ -73,7 +74,7 @@ exports.enable = function (Module) {
     var countBefore = entry && entry.runCount;
     var exports = typeof module.require === "function"
       ? module.require(absoluteId)
-      : require(absoluteId);
+      : dynRequire(absoluteId);
 
     if (entry && entry.runCount === countBefore) {
       // If require(absoluteId) didn't run any setters for this entry,
--- a/node/caching-compiler.js
+++ b/node/caching-compiler.js
@@ -2,6 +2,7 @@ var fs = require("fs");
 var path = require("path");
 var createHash = require("crypto").createHash;
 var compile = require("../lib/compiler.js").compile;
+var dynRequire = module.require ? module.require.bind(module) : __non_webpack_require__;
 var hasOwn = Object.prototype.hasOwnProperty;
 
 // Map from absolute file paths to the package.json that governs them.
@@ -70,7 +71,7 @@ function readWithCache(info, content) {
   };
 
   if (reify && reify.parser) {
-    compileOptions.parse = require(reify.parser).parse;
+    compileOptions.parse = dynRequire(reify.parser).parse;
   };
 
   content = compile(content, compileOptions).code;
--- a/node/runtime.js
+++ b/node/runtime.js
@@ -1,4 +1,5 @@
-var Module = require("../lib/runtime.js").enable(module.constructor);
+var dynRequire = module.require ? module.require.bind(module) : __non_webpack_require__;
+var Module = require("../lib/runtime.js").enable(dynRequire('module'));
 var Mp = Module.prototype;
 
 exports.Module = Module;


lib/compiler.js
```diff
@@ -1,9 +1,13 @@
...
+var types = require("ast-types/fork")([
+  require("ast-types/def/esprima"),
+  require("ast-types/def/babel6")
+]);
+
 var n = types.namedTypes;
 var b = types.builders;
-var MagicString = require("magic-string");
+var MagicString = require("magic-string/dist/magic-string.cjs.js");
 var utils = require("./utils.js");
```
jdalton: I noticed that by default `magic-string/dist/magic-string.es.js` is loaded. I think that's because by this time the loader is already registered and so the es flavor is loaded.

lib/compiler.js
```diff
@@ -1,9 +1,13 @@
 var hasOwn = Object.prototype.hasOwnProperty;
 var assert = require("assert");
-var types = require("ast-types");
+var types = require("ast-types/fork")([
+  require("ast-types/def/esprima"),
+  require("ast-types/def/babel6")
+]);
```
jdalton: esprima loads es7, and es7 loads es6, and es6 loads core.
babel6 loads babel and flow.

lib/parsers/default.js
```diff
@@ -3,14 +3,14 @@ var REIFY_PARSER =
...
+    REIFY_PARSER === "babylon" ||
+    REIFY_PARSER === "top-level") {
+  exports.parse = require("./" + process.env.REIFY_PARSER + ".js").parse;
 } else try {
   exports.parse = require(REIFY_PARSER);
 } catch (e) {
-  exports.parse = require("./acorn.js").parse;
+  var id = "./acorn.js";
+  exports.parse = require(id).parse;
 }
```
jdalton: This tweak allows me to use:
```js
new webpack.DefinePlugin({
   'process.env.REIFY_PARSER': JSON.stringify('top-level')
})
```

lib/runtime.js
```diff
@@ -1,6 +1,11 @@
 var Entry = require("./entry.js").Entry;
 var utils = require("./utils.js");
 
+// Support dynamic require within webpack bundle.
+if (typeof __non_webpack_require__ === 'undefined') {
+  var __non_webpack_require__ = require;
+}
```
jdalton: The hoisted `var` is a bit gross but allows webpack to work its magic.

jdalton:
@RReverser any tips for bundling a smaller acorn package? I can get the bundle size down to 40 kB min+zopfli (which is great, but wondering if it could be even smaller)

RReverser:
@jdalton Is this source-map-explorer output?

node/caching-compiler.js
```diff
@@ -70,7 +71,7 @@ function readWithCache(info, content) {
   };
 
   if (reify && reify.parser) {
-    compileOptions.parse = require(reify.parser).parse;
+    compileOptions.parse = utils.require(reify.parser).parse;
```
benjamn: Come to think of it, could we just use `module.require` here (and other places where `utils.require` is used)? That way, module resolution would be relative to `module.id` rather than always `lib/utils.js`.

lib/utils.js
```diff
@@ -7,6 +7,12 @@ function isPlainObject(value) {
 }
 exports.isPlainObject = isPlainObject;
 
+// Support dynamic require within webpack bundle.
+if (typeof __non_webpack_require__ === 'undefined') {
+  var __non_webpack_require__ = require;
```
benjamn: Hmm, I think the presence of this `var` is going to ensure that `__non_webpack_require__ === 'undefined'` is always true. How about this:
```js
exports.require = typeof __non_webpack_require__ === 'undefined'
  ? require : __non_webpack_require__;
```

lib/compiler.js
benjamn: Calling `require("magic-string/dist/magic-string.cjs.js")` makes perfect sense to me!

lib/compiler.js
benjamn: As we chatted about yesterday, the longer term plan is to avoid using the `ast-types` library if we can get away with something lighter-weight.

lib/utils.js
RReverser: @benjamn I don't think so. `var` hoisting means only that variable is already set to `undefined` when it's accessed, basically same as 

```javascript
let __non_webpack_require__ = undefined;
if (typeof __non_webpack_require__ === 'undefined') {
  __non_webpack_require__ = require;
}
```

just shorter with ES5 `var`.

benjamn: var __non

benjamn: What I'm saying is that the `typeof` check isn't doing what it's supposed to, thanks to the `var` binding:
```js
var value = 123;
console.log(function() {
  if (typeof value === "undefined") {
    var value = 456;
  }
  return value;
}()); // logs 456
```

jdalton: > Hmm, I think the presence of this var is going to ensure that __non_webpack_require__ === 'undefined' is always true.

You're right for running code it's always `undefined`. Which is what I want since this is a hint for the webpack bundler when it's running its bundle transform. When the bundle transform happens only the invocations are changed. So `___non_webpack_require__(smth)` is changed to `require(smth)`. If you try to set the variable outside of the block then it loses its webpack magic-ness. I've seen others set it on the `global` property but I wanted to avoid that and after some tinkering found that I could get it to work by using a var in a block.

jdalton: So utils.js is transformed from:
```js
// Support dynamic require within webpack bundle.
if (typeof __non_webpack_require__ === 'undefined') {
  var __non_webpack_require__ = require;
}
exports.require = require;
```

```
to
```js
// Support dynamic require within webpack bundle.
if (false) {
  var __non_webpack_require__ = require;
}
exports.require = require;
```

jdalton: Ah nice! @benjamn because you're take is assigning it directly t `exports.require` it gets transformed into:
```js

// Support dynamic require within webpack bundle.
exports.require =  false
  ? require : require;
```
which works! I had run into issues where webpack would create a local undefined var of `require` in some cases.

jdalton:
@RReverser 
>  Is this source-map-explorer output? Asking because it looks like output of this new webpack tool, which just shows what's in the bundle, but does not try to show files scaled to their real size, so it's useless for size measurements and reduction.

Acorn alone is 30kb min+gzipped which is 75% of the overall bundle size.

node/caching-compiler.js
jdalton: Ah each require is per instance of the module so it could effect resolution. There is no `module.require` in webpack transformed code so I might have to move the transform per file (that's how I had it before I tried to consolidate).

lib/compiler.js
jdalton: I think it's OK for now as the main weight comes from acorn. Let's move ast-types talk to https://github.com/benjamn/ast-types/pull/209.

jdalton:
@RReverser since acorn is a large chunk of the bundle any pointers to creating a custom build or any insights into lesser used bits of acorn that could be omitted from a custom build would help.

lib/runtime.js
```diff
@@ -1,5 +1,6 @@
 var Entry = require("./entry.js").Entry;
 var utils = require("./utils.js");
+var dynRequire = module.require || __non_webpack_require__;
 
 exports.enable = function (Module) {
```
jdalton: Fixed the `require` call by simplifying it and making it local per file.

jdalton: This
```js
var dynRequire = module.require || __non_webpack_require__
``
is transformed into
```js
var dynRequire = module.require || require
```
Since the webpack bundled form doesn't have `module.require` it falls back to the real `require`.

RReverser:
> Since acorn is a large chunk of the bundle any pointers to creating a custom build or any insights into lesser used bits of acorn that could be omitted from a custom build would help.

I'm afraid that would be quite a difficult task as we're still parsing top-level statements (which means basically any syntax).

jdalton:
> I'm afraid that would be quite a difficult task as we're still parsing top-level statements (which means basically any syntax, so not much to strip).

Ok that's cool. As I mentioned above 40 kB is fantastic. I was just curious if there was anything else to strip away. Thank you!

lib/runtime.js
benjamn: This is looking good, just one more issue: `module.require` is an unbound function that refers to `this` ([here](https://github.com/nodejs/node/blob/5bda5faffdeefa448965159e524eacdd14490436/lib/module.js#L513)), so I think this needs to be
```js
var dynRequire = module.require ? module.require.bind(module) : require;
```

jdalton: Cool, I'll update. Is there a unit test to check that as currently they're passing.

jdalton: Updated.

PR #78, opened 2017-03-18
Use `Object.create(null)` in a few more places.

jdalton:
I noticed while trying to load `lodash-es` that an error occured [here](https://github.com/benjamn/reify/blob/ae4c671e76f78c200b523ed0f94343084f3c8f73/lib/compiler.js#L653-L655) because I had an identifier named `toString` and the `map` as a vanilla object and so had a `toString` method and not the array it was expecting:
```js
var locals = map[__ported] || [];
```

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -490,7 +490,7 @@ var importExportVisitor = new types.PathVisitor.fromMethodsObject({
     var dd = decl.declaration;
 
     if (dd) {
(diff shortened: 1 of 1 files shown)


jdalton:
This should totally have a unit test but I'm not super well versed yet with the setup so it may be easier for you to add in a subsequent commit.

PR #85, opened 2017-03-20
Rename Module.prototype.import to Module.prototype.importSync.

### LINKED ISSUE #84: Make way for async import.
The esm built-in `import()` will be async. I :heart: the `module.import` namespace as a userland alternative. The current `module.import` is synchronous so I propose renaming `module.import` to `module.importSync` *(similar to the \*Sync forms of other Node built-in methods)* leaving room for `module.import` to be the async form to align more closely with built-in `import()`.

benjamn:
Because Reify controls both the runtime and the compiler, I believe this change can be made in one commit without breaking backwards compatibility, though it definitely deserves a minor version bump.

Closes #84.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,12 +39,12 @@ examples. While you do not have to write this API by hand, it is designed
 to be easily human readable and writable, in part because that makes it
 easier to explain.
 
-I will explain the `Module.prototype.import` method first, then the
+I will explain the `Module.prototype.importSync` method first, then the
 `Module.prototype.export` method after that. Note that this `Module` is
 the constructor of the CommonJS `module` object, and the `import` and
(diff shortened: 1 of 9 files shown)


jdalton:
Looks good 👓

README.md
```diff
@@ -67,11 +67,11 @@ module.import("./module", {
 });
 ```
 
-All setter functions are called synchronously before `module.import`
-returns, with whatever values are immediately available. However, when
-there are import cycles, some setter functions may be called again, when
-the exported values change. Calling these setter functions one or more
-times is the key to implementing [*live
+All setter functions are called synchronously before `module.importSync`
+returns (hence the `importSync` name), with whatever values are
```
jdalton: 👏

Issue #86, opened 2017-03-20
Improve performance of ast-types.


jdalton:
I wanted to bubble up the issue with [Reify's use](https://github.com/benjamn/reify/blob/4eb7442dc45ac786663036913163bf088529981b/lib/compiler.js#L3-L6) of [ast-types](https://github.com/benjamn/ast-types) so that folks like @hzoo (or someone he may recommend) can help.

Related to #79.

benjamn:
Besides performance, removing the `ast-types` dependency would liberate Reify from having to care about the exact set of AST types the parser is using (Esprima, Babylon, etc.), as long as the import/export-related types agree.

PR #87, opened 2017-03-20
Stop using ast-types.

benjamn:
Using a custom `Visitor` abstraction instead of the [`PathVisitor`](https://github.com/benjamn/ast-types/blob/master/lib/path-visitor.js) API provided by `ast-types` is less convenient, but it eliminates the second-largest dependency of this package (after `acorn`), and opens the door to Reify-specific optimizations that might not be safe for a more generic API.

This PR should help with the goals of #86 (optimizing compiler performance), even though it removes `ast-types` from the picture. For example, we no longer have to import a bunch of `ast-types` definitions when we import the compiler for the first time.

Another benefit is that we no longer have to worry about variations in AST type hierarchies generated by different parsers, because this visitor only cares about `import`/`export`-related nodes and tolerates/ignores everything else.

In my tests this abstraction is just as fast as, if not faster than, the old one.

cc @jdalton

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const assert = require("assert");
+const undefined = void 0;
 
 // A simplified version of the FastPath AST traversal abstraction used by
 // Recast: https://github.com/benjamn/recast/blob/master/lib/fast-path.js
@@ -101,24 +102,13 @@ FPp.valueIsNode = function () {
 };
 
 FPp.getNode = function getNode(count) {
-  return getNodeHelper(this, ~~count);
+  if (count === undefined) count = 0;
+  return getNodeHelper(this, count);
 };
 
 FPp.getParentNode = function getParentNode(count) {
-  return getNodeHelper(this, ~~count + 1);
-};
-
-// The length of the stack can be either even or odd, depending on whether
-// or not we have a name for the root value. The difference between the
-// index of the root value and the index of the final value is always
-// even, though, which allows us to return the root value in constant time
-// (i.e. without iterating backwards through the stack).
-FPp.getRootValue = function getRootValue() {
-  const s = this.stack;
-  if (s.length % 2 === 0) {
-    return s[1];
-  }
-  return s[0];
+  if (count === undefined) count = 0;
+  return getNodeHelper(this, count + 1);
 };
 
 // Temporarily push properties named by string arguments given after the
--- a/lib/visitor.js
+++ b/lib/visitor.js
@@ -1,6 +1,7 @@
 "use strict";
 
 const hasOwn = Object.prototype.hasOwnProperty;
+const underscoreCode = "_".charCodeAt(0);
 
 // This Visitor API was inspired by a similar API provided by ast-types:
 // https://github.com/benjamn/ast-types/blob/master/lib/path-visitor.js
@@ -8,7 +9,7 @@ const hasOwn = Object.prototype.hasOwnProperty;
 function Visitor() {
   this.visit = this.visit.bind(this);
   this.visitWithoutReset = this.visitWithoutReset.bind(this);
-  this.traverse = this.traverse.bind(this);
+  this.visitChildren = this.visitChildren.bind(this);
 }
 
 module.exports = Visitor;
@@ -26,36 +27,39 @@ Vp.visitWithoutReset = function (path) {
   } else if (path.getNode() === value) {
     const method = this["visit" + value.type];
     if (typeof method === "function") {
-      // Note: the method must call this.traverse(path) to continue
+      // Note: the method must call this.visitChildren(path) to continue
       // traversing.
       method.call(this, path);
     } else {
-      this.traverse(path);
+      this.visitChildren(path);
     }
   }
 };
 
-Vp.traverse = function (path) {
+Vp.visitChildren = function (path) {
   if (! path.valueIsNode()) {
     return;
   }
 
   const node = path.getValue();
+  const keys = Object.keys(node);
+  const keyCount = keys.length;
 
-  Object.keys(node).forEach(function (key) {
-    if (key === "loc") {
-      return;
-    }
+  for (let i = 0; i < keyCount; ++i) {
+    const key = keys[i];
 
-    if (key.charAt(0) === "_") {
-      return;
+    if (key === "loc" || // Ignore .loc.{start,end} objects.
+        // Ignore "private" properties added by Babel.
+        key.charCodeAt(0) === underscoreCode) {
+      continue;
     }
 
     const value = node[key];
     if (! (value && typeof value === "object")) {
-      return;
+      // Ignore properties whose values aren't objects.
+      continue;
     }
 
     path.call(this.visitWithoutReset, key);
-  }, this);
+  }
 };
--- a/lib/assignment-visitor.js
+++ b/lib/assignment-visitor.js
@@ -29,7 +29,8 @@ AVp.visitUpdateExpression = function (path) {
 };
 
 AVp.visitCallExpression = function (path) {
-  this.traverse(path);
+  this.visitChildren(path);
+
   var callee = path.getValue().callee;
   if (callee.type === "Identifier" &&
       callee.name === "eval") {
@@ -38,7 +39,7 @@ AVp.visitCallExpression = function (path) {
 };
 
 AVp._assignmentHelper = function (path, childName) {
-  this.traverse(path);
+  this.visitChildren(path);
 
   var child = path.getValue()[childName];
   var assignedNames = utils.getNamesFromPattern(child);


jdalton:
Woah, I can't wait to dig into this in the evening 😮

lib/visitor.js
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    if (key.charAt(0) === "_") {
+      return;
+    }
+
+    const value = node[key];
+    if (! (value && typeof value === "object")) {
+      return;
+    }
```
benjamn: @jdalton In my experience, skipping keys in this kind of loop can have huge traversal performance benefits, if you can think of any other kinds of AST properties that we could ignore.

lib/compiler.js
```diff
@@ -187,13 +186,13 @@ var importExportVisitor = new types.PathVisitor.fromMethodsObject({
...
 
-      if (n.SwitchCase.check(stmtPath.value)) {
+      if (stmtPath.value.type === "SwitchCase") {
         assert.strictEqual(bodyPath.name, "consequent");
         insertCharIndex = bodyPath.value[0].start;
         break;
       }
 
-      if (n.Statement.check(stmtPath.value)) {
+      if (! Array.isArray(stmtPath.value)) {
```
jdalton: Just so I understand better can you point to where `Statement.check` is equiv to `Array.isArray`.

lib/utils.js
```diff
@@ -29,3 +29,52 @@ exports.shallowObjEqual = function(a, b) {
...
+    case "ObjectPattern":
+      queue.push.apply(queue, pattern.properties);
+      break;
+    case "ArrayPattern":
+      queue.push.apply(queue, pattern.elements);
+      break;
+    case "RestElement":
+      queue.push(pattern.argument);
+      break;
+    }
```
jdalton: What does the compiler use the `ArrayPattern` or `RestElement` for? It's not `import`/`export` related is it?

lib/fast-path.js
```diff
@@ -0,0 +1,202 @@
...
+const codeOfA = "A".charCodeAt(0);
+const codeOfZ = "Z".charCodeAt(0);
+function isCapitalized(string) {
+  if (typeof string !== "string") return false;
+  const code = string.charCodeAt(0);
+  return code >= codeOfA && code <= codeOfZ;
+}
+
+FPp.getNode = function getNode(count) {
+  return getNodeHelper(this, ~~count);
```
jdalton: I always forget what the `~~` is doing. Is it a cheap `Math.floor`?
I know it has range limits more verbose forms don't.

jdalton: For example `~~Math.pow(2,32) // => 0`

jdalton: Is `count` something that needs to be floored?

lib/fast-path.js
```diff
@@ -0,0 +1,202 @@
...
+// or not we have a name for the root value. The difference between the
+// index of the root value and the index of the final value is always
+// even, though, which allows us to return the root value in constant time
+// (i.e. without iterating backwards through the stack).
+FPp.getRootValue = function getRootValue() {
+  const s = this.stack;
+  if (s.length % 2 === 0) {
+    return s[1];
+  }
+  return s[0];
```
jdalton: If you reversed the check it could just be a truthy check.

lib/visitor.js
```diff
@@ -0,0 +1,39 @@
...
+    } else {
+      this.traverse(path);
+    }
+  }
+};
+
+Vp.traverse = function (path) {
+  Object.keys(path.getValue()).forEach(function (key) {
+    path.call(this.visitWithoutReset, key);
+  }, this);
```
jdalton: How often is this called? `forEach` may have probs being inlined and then you throw on top of it a bind... if it's potentially hundreds and thousands change it to a for-loop and the `this` is free.

lib/visitor.js
```diff
@@ -33,7 +33,26 @@ Vp.visitWithoutReset = function (path) {
...
+  }
+
+  const node = path.getValue();
+
+  Object.keys(node).forEach(function (key) {
+    if (key === "loc") {
+      return;
+    }
+
+    if (key.charAt(0) === "_") {
```
jdalton: charCode is a faster compare

lib/visitor.js
jdalton: I'm not sure what you're looking for so not sure what to ignore.

lib/visitor.js
```diff
@@ -33,7 +33,26 @@ Vp.visitWithoutReset = function (path) {
...
 
 Vp.traverse = function (path) {
-  Object.keys(path.getValue()).forEach(function (key) {
+  if (! path.valueIsNode()) {
+    return;
+  }
+
+  const node = path.getValue();
+
+  Object.keys(node).forEach(function (key) {
```
jdalton: That for loop-advice from early fits here.

lib/visitor.js
benjamn: With `ast-types`, we know exactly what fields a given AST node is supposed to have, based on its `.type` (see [here](https://github.com/benjamn/ast-types/blob/e855515735dd84e415dd03319e0cd6e1ecb59204/lib/path-visitor.js#L223)). Because we don't know anything about the expected structure of AST nodes, there's a chance here that we're visiting properties of nodes that don't matter. For example, the `.loc.{start,end}` object isn't going to matter for this transform, and Babel attaches "private" properties (containing metadata, not AST nodes) by prefixing their names with `_`.

My general point is just that any intelligent short cuts we can take to ignore whole subtrees of the AST could pay off big time. Like, if we knew whether or not a subtree contained any `import` or `export` tokens, that might allow us to skip lots of subtrees. That kind of thing.

lib/compiler.js
benjamn: It's definitely not equivalent in general, but this change works because an `import` declaration can only be contained by a few kinds of nodes:
1. the `.body` array of a `Program`
2. the `.body` array of a `BlockStatement`
3. the `.consequent` array of a `SwitchCase`
4. a `Statement` without braces, such as:
```js
if (loggingEnabled) import "logging";
```
The general `Statement.check` API is something we gave up by removing `ast-types`, since it requires knowledge of all subtypes of `Statement`, which we don't have any more. But that's OK because we can still identify case 4 because it's the only case where this loop encounters an AST node before it encounters a `.body` array.

Of course, if you're using the top-level parser, this logic won't be exercised (in fact only the `Program` case should ever be reached), but Reify is intended to handle nested `import` declarations if you give it a full AST.

lib/utils.js
benjamn: It's part of the `module.runModuleSetters` wrapping that `AssignmentVisitor` does.

Consider this code:
```js
export let a, rest;

// Elsewhere in the file:
[a, ...rest] = foo;
```

This becomes
```js
module.export({
  a: function() { return a },
  rest: function() { return rest }
});
let a, rest;

// Elsewhere in the file:
module.runModuleSetters([a, ...rest] = foo);
```
Because `a` and `rest` are exported, the `AssignmentVisitor` uses `getNamesFromPattern` to find places where they are reassigned, and wraps those `AssignmentExpression`s with a call to `module.runModuleSetters`.

Now, if we could be sure that the assignment happened before the module finished loading for the first time, then we could avoid the wrapping, because `module.runModuleSetters` is idempotent and we call it automatically when a module finishes loading. In general, though, Reify has to worry about reassignments that happen long after the module finishes loading.

It's tempting to disable the `AssignmentVisitor` when using the top-level parser, but I haven't fully thought through the consequences of that.

It's also worth mentioning that the top-level parsers sacrifices spec compliance by not updating exports that get reassigned in nested scopes. It's worth thinking about whether that's really OK, but it's also entirely possible for the library author (e.g. you) to call `module.runModuleSetters()` manually in cases where updating the exported value is really important. If that's too much to ask, the developer should not use the top-level parser. With clever optimizations comes additional responsibility for correctness.

lib/fast-path.js
benjamn: Yeah, this is just inherited from Recast. It's just coercing `undefined` to `0`, which can be accomplished lots of better ways.

lib/fast-path.js
benjamn: Truth!

lib/visitor.js
benjamn: Once for (nearly) every node in the (top-level) AST. I'll see if a `for` loop helps.

lib/fast-path.js
benjamn: Even cuter: `return s[(s.length + 1) % 2]`

Though I appreciate the feedback, I think I'm actually going to remove this method from the file for now, since it isn't used yet.

lib/utils.js
jdalton: We'll have to go over `runModuleSetters` on the call. I've looked into it and can't figure out how it does what it does.

jdalton: I'm wondering if there's any possible gaps. Also, if it could be optimized to only track primitive values.

benjamn: Because I really didn't want to have to explain how that method works, I took some time this morning to simplify it considerably: https://github.com/benjamn/reify/pull/89

jdalton:
This PR will need to be rebased but then it gets :+1:

PR #88, opened 2017-03-21
Write cache files to disk asynchronously.

benjamn:
Judging from the results of https://github.com/jdalton/reify-perf-check, this noticeably reduces the overhead from caching.

Before:
```sh
~/dev/reify-perf-check% npm test

> perf-example@0.1.0 test /Users/ben/dev/reify-perf-check
> node index

NO REIFY CACHE
loading cjs: 204.566ms
loading reify: 46.905ms
loading cjs after reify: 149.188ms
loading esm: 835.038ms
loading esm.gz: 799.494ms

REIFY CACHED
loading cjs: 110.203ms
loading esm: 248.458ms
loading esm.gz: 330.240ms
```

After:
```sh
~/dev/reify-perf-check% npm test

> perf-example@0.1.0 test /Users/ben/dev/reify-perf-check
> node index

NO REIFY CACHE
loading cjs: 196.121ms
loading reify: 44.745ms
loading cjs after reify: 142.679ms
loading esm: 709.120ms
loading esm.gz: 700.499ms

REIFY CACHED
loading cjs: 78.973ms
loading esm: 240.221ms
loading esm.gz: 331.046ms
```

### DIFF
--- a/node/caching-compiler.js
+++ b/node/caching-compiler.js
@@ -97,13 +97,25 @@ function compileWithCache(pkgInfo, content, options) {
   pkgInfo.cache[cacheFilename] = content;
 
   if (typeof pkgInfo.cacheDir === "string") {
-    // Writing cache files is something that should only happen at package
-    // development time, so it's acceptable to use fs.writeFileSync
-    // instead of some complicated asynchronous-but-atomic strategy.
-    fs.writeFileSync(absCachePath, content, "utf8");
+    scheduleWrite(absCachePath, content);
   }
 
   return content;
+}
+
+const pendingWrites = Object.create(null);
+let pendingWriteTimer;
+
+function scheduleWrite(path, content) {
+  pendingWrites[path] = content;
+  pendingWriteTimer = pendingWriteTimer || setTimeout(function () {
+    pendingWriteTimer = null;
+    Object.keys(pendingWrites).forEach(function (path) {
+      const content = pendingWrites[path];
+      delete pendingWrites[path];
+      fs.writeFileSync(path, content, "utf8");
+    });
+  }, 10);
 }
 
 function readFileOrNull(filename) {


jdalton:
Ah nice! I dig that revisiting smth once other optimizations are in place shows the impact more. Woot! Btw the minified+zopflied bundle is down to 30 kB.

node/caching-compiler.js
```diff
@@ -84,15 +86,27 @@ function readWithCache(info, content) {
...
-    // development time, so it's acceptable to use fs.writeFileSync
-    // instead of some complicated asynchronous-but-atomic strategy.
-    fs.writeFileSync(absCachePath, content, "utf8");
+    scheduleWrite(absCachePath, content);
   }
 
   return content;
 }
 
+const pendingWrites = Object.create(null);
```
jdalton: This looks like a place for a `Map` to begin dipping a toe into that.

jdalton: Actually looking at it more couldn't this just be done with an array of paths to write?

benjamn: Sure, but an object makes it easy to store file contents as property values.

jdalton: NM. Ah, I missed you were using both the path and the content.

node/caching-compiler.js
```diff
@@ -84,15 +86,27 @@ function readWithCache(info, content) {
...
+  pendingWrites[path] = content;
+  pendingWriteTimer = pendingWriteTimer || setTimeout(function () {
+    pendingWriteTimer = null;
+    Object.keys(pendingWrites).forEach(function (path) {
+      const content = pendingWrites[path];
+      delete pendingWrites[path];
+      fs.writeFileSync(path, content, "utf8");
+    });
+  }, 10);
+}
```
jdalton: When I first saw the PR title I thought each write would be async but this is a batch of them.
Are there cases where this could get into a cache jam? Could it be writing to a cache file that was written at some point before or could an older cache get written trumping a newer one (some kind of race condition).

benjamn: This line
```js
pendingWrites[path] = content;
```
*could* technically overwrite an existing path/content entry, but I would hope the `content` would be the same, and newer content would be preferable anyway if the content was different.

My experience with writing cache files asynchronously is that they can end up corrupted if the app is killed mid-write (often an empty file is created but never written to), and that's tricky to detect when reading from the cache, since these are just `.js` files that very well could be empty.

We could potentially break up the flush, limiting it to a certain number of writes, or a certain time limit per flush, and just reschedule for 10ms later. I think that would be safe enough. Any thoughts on that?

jdalton: How are cache names generated is it based on full file paths?

jdalton: > that's tricky to detect when reading from the cache, since these are just .js files that very well could be empty.

Don't they currently require an import or export mention to be processed? If so and empty files are a thing then that can be a check.

jdalton: Can you use `setImmediate`?

benjamn: > How are cache names generated. Is it based on full file paths?

It's a [hash](https://github.com/benjamn/reify/blob/478538bc9d6bc2cd2dde8a07f0de6cd5000314da/node/caching-compiler.js#L104) of file contents and the "reify" options from `package.json`, so I'm pretty sure it's bulletproof (if you aren't worried about hash collisions).

benjamn: > Can you use `setImmediate`?

It's not the end of the world if some cache files don't get written immediately, so I'm not too concerned about making it happen sooner. Is there some other reason for using `setImmediate`?

jdalton: > It's a hash of file contents and the "reify" options from package.json

Awesome, I was hoping that was it.

> Is there some other reason for using `setImmediate`?

If there's no compat concern it's the "more correct" way to defer an operation is all.

jdalton: Have you experimented with how expensive hashing the file is? I'm curious if that could be a pain point when multiplied by hundreds of modules.

node/caching-compiler.js
jdalton: Then a `Map` might work well.

node/caching-compiler.js
benjamn: We chatted about this offline yesterday, but for anyone else who might be reading: hashing large files definitely has a cost, and we agreed that just hashing the package version and the file path should be adequate in production.

jdalton: Maybe the `mtime` too.

jdalton:
I think once this PR is rebased it's fine to go ahead and test the waters with it. Worse case we adjust a bit for edge cases.

benjamn:
Rebased!

PR #89, opened 2017-03-23
Simplify logic for calling runModuleSetters for parent modules.

benjamn:
When @jdalton tells you he can't make sense of your code, it's probably worth simplifying.

Trying to detect whether setters modified `module.exports` objects of any parent modules was never strictly necessary, since it was always safe to call `runModuleSetters` blindly and trust the `setter.last` cache to prevent the broadcast from falling into an infinite loop.

Whatever the performance benefits of avoiding calling `runModuleSetters` might have been, the benefits of not taking snapshots of parent `module.exports` objects are probably at least as significant.

### DIFF
--- a/lib/entry.js
+++ b/lib/entry.js
@@ -85,38 +85,25 @@ function runModuleGetters(module) {
 
 Ep.runModuleGetters = function (module) {
   var entry = this;
-  var changeCount = 0;
-
   Object.keys(entry.getters).forEach(function (name) {
-    if (entry.runGetter(module, name)) {
-      ++changeCount;
-    }
+    entry.runGetter(module, name);
   });
-
-  return changeCount;
 };
 
 // Returns true iff the getter updated module.exports with a new value.
 Ep.runGetter = function (module, name) {
-  if (! hasOwn.call(this.getters, name)) {
-    return false;
+  if (hasOwn.call(this.getters, name)) {
+    try {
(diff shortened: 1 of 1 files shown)


lib/entry.js
```diff
@@ -170,75 +157,34 @@ Ep.runModuleSetters = function (module) {
...
-    // If we never called takeSnapshot, then we can avoid checking
-    // relevantParents and parentSnapshots below.
     return;
   }
 
   // If any of the setters updated the module.exports of a parent module,
   // or updated local variables that are exported by that parent module,
   // then we must re-run any setters registered by that parent module.
   Object.keys(relevantParents).forEach(function (id) {
-    var parent = relevantParents[id];
```
jdalton: I'd caution use of  `Array#forEach` and other ES5/6 functional array helpers in code that you'd like to be snappy. One off calls are OK but when you stack them they can bog down execution *(engines haven't done a great job of optimizing them even 12 yrs later)*. With `let`, `const`, and `for-of` it allows many of the readability and logic-guard wins of `Array#forEach` and friends with less cumulative overhead.

benjamn: Just pushed a commit to use `let` and `const` and avoid `Object.keys(...).forEach` in `lib/entry.js`.

jdalton:
👍

benjamn:
:tada:

PR #91, opened 2017-03-24
Make let generation the default.

jdalton:
Now that Node 4 is the min bar let by default would be rad. For the case of the REPL I'm guessing it could be configured `false` for REPL use. I'm missing the option toggle for the REPL use.

### DIFF
--- a/repl/index.js
+++ b/repl/index.js
@@ -17,7 +17,7 @@ function wrap(name, optionsArgIndex) {
   vm[name] = function (code) {
     var options = arguments[optionsArgIndex];
     var filename = options && options.filename;
-    var args = [compile(code, filename)];
+    var args = [compile(code, filename, { generateLetDeclarations: false })];
     for (var i = 1; i < arguments.length; ++i) {
       args.push(arguments[i]);
     }
--- a/lib/options.js
+++ b/lib/options.js
@@ -3,7 +3,7 @@
 const hasOwn = Object.prototype.hasOwnProperty;
 
 const defaultCompileOptions = {
-  generateLetDeclarations: false,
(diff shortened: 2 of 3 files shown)


jdalton:
I'm missing the option toggle for the REPL use.

jdalton:
While working on this PR I noticed some nested import tests failing
```js
switch (i) {
case 0:
  import { a as x } from "./abc";
  break;
case 1:
  import { b as x } from "./abc";
  break;
}
```
because:
```
SyntaxError: Identifier 'x' has already been declared
```
which makes sense because it's `let`. This is something to keep in mind as a toggle for nested import statements.

repl/index.js
```diff
@@ -17,7 +17,7 @@ function wrap(name, optionsArgIndex) {
   vm[name] = function (code) {
     var options = arguments[optionsArgIndex];
     var filename = options && options.filename;
-    var args = [compile(code, filename)];
+    var args = [compile(code, filename, { generateLetDeclarations: false })];
     for (var i = 1; i < arguments.length; ++i) {
```
jdalton: Being able to set options for the REPL is probably a good thing regardless of the current nested import `let` issue.

benjamn:
Re: `switch` case scope collisions, I'm tempted to wrap the `case:` bodies with braces, though I guess that would be an observable semantic change. Where should the `let` declarations go, if we wanted to make your example work?

jdalton:
> Where should the let declarations go, if we wanted to make your example work?

If possible it could work by using `let` for top level import and `var` for nested.

jdalton:
I'll hold off on this PR until a few other things are in place.

PR #94, opened 2017-03-27
Use more ES6 in node code.

jdalton:
This PR uses more ES6 in node code.

### DIFF
--- a/node/caching-compiler.js
+++ b/node/caching-compiler.js
@@ -1,27 +1,29 @@
-var fs = require("fs");
-var path = require("path");
-var createHash = require("crypto").createHash;
-var compile = require("../lib/compiler.js").compile;
-var dynRequire = module.require ? module.require.bind(module) : __non_webpack_require__;
-var hasOwn = Object.prototype.hasOwnProperty;
+"use strict";
+
+const fs = require("fs");
+const path = require("path");
+const createHash = require("crypto").createHash;
+const compile = require("../lib/compiler.js").compile;
+const dynRequire = module.require ? module.require.bind(module) : __non_webpack_require__;
+const hasOwn = Object.prototype.hasOwnProperty;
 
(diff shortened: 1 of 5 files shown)


repl/index.js
```diff
@@ -1,24 +1,23 @@
...
-    return;
-  }
-
-  if (method.reified) {
+  if (typeof method !== "function" ||
+      method.reified) {
     return;
   }
 
   vm[name] = function (code) {
```
jdalton: Interesting that if I make, `vm[name] = (code) => {`, an arrow function tests fail.
Do you know if it's inspecting the toString of these functions or if they require them to have a `.prototype`?

benjamn: I'll investigate.

node/caching-compiler.js
```diff
@@ -162,17 +166,16 @@ function readPkgInfo(dir) {
...
       // reification even if "reify" is listed as a dependency.
       return null;
     }
 
-    function check(name) {
-      return typeof pkg[name] === "object" &&
-        hasOwn.call(pkg[name], "reify");
-    }
+    const check = (name) => (
+      typeof pkg[name] === "object" && hasOwn.call(pkg[name], "reify")
```
benjamn: Note to self: add a `;` here.

jdalton: Yes! On my todo is to add some linting.

repl/index.js
jdalton: Ya, it's not critical to this PR just curious (seems interesting).

node/caching-compiler.js
benjamn: Oh wait, that's an expression-bodied arrow function. Never mind!

repl/index.js
jdalton: Ah, it's because arrow functions don't have `arguments` objects so it's referencing the `arguments` object of the module itself.

PR #95, opened 2017-03-27
Allow setting compiler options in node/index.js.

jdalton:
This PR is a take at allowing the node/index.js to accept compiler option overrides.

It allows:
```js
require("reify/node")({
  someSetting: true,
  otherSetting: "path/to/smth"
})
```

The interface could be different though, this is just an initial take on it.

Related to #91.

### DIFF
--- a/node/compile-hook.js
+++ b/node/compile-hook.js
@@ -5,6 +5,12 @@ const compiler = require("./caching-compiler.js");
 const Module = require("./runtime.js").Module;
 const Mp = Module.prototype;
 
+let overrideCompileOptions = {};
+
+module.exports = exports = (options) => {
+  overrideCompileOptions = Object.assign({}, options);
+};
+
 // Override Module.prototype._compile to compile any code that will be
 // evaluated as a module.
 const _compile = Mp._compile;
@@ -12,7 +18,7 @@ if (! _compile.reified) {
   (Mp._compile = function (content, filename) {
     return _compile.call(
       this,
-      compiler.compile(content, {
+      compiler.compile(content, Object.assign({
         filename: filename,
         makeCacheKey() {
           const stat = compiler.statOrNull(filename);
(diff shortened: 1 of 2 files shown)


jdalton:
It allows:
```js
require("reify/node")({
  someSetting: true,
  otherSetting: "path/to/smth"
})
```

The interface could be different though, this is just an initial take on it.

node/caching-compiler.js
```diff
@@ -48,7 +48,7 @@ const fallbackPkgInfo = {
   cache: Object.create(null)
 };
 
-function readWithCache(info, content) {
+function readWithCache(info, content, options) {
   const json = info && info.json;
   const reify = json && json.reify;
   const cacheFilename = getCacheFilename(content, reify);
```
benjamn: The new options should probably be included in the hash computed by `getCacheFilename`, right?

jdalton: I don' think `getCacheFilename` is using any options at the moment.

benjamn: But the options might affect the compiler output, and we want `getCacheFilename` to return different names for different versions of the compiled code.

jdalton:
I'll try to update this one tonight.

jdalton:
Rebased.

node/compile-hook.js
```diff
@@ -5,14 +5,20 @@ const compiler = require("./caching-compiler.js");
 const Module = require("./runtime.js").Module;
 const Mp = Module.prototype;
 
+let overrideCompileOptions = {};
+
+module.exports = (options) => {
```
benjamn: Maybe also update the `exports` variable in this assignment, just in case we ever accidentally use that in this module.

jdalton: I'm not sure what you're asking. Can you give an example.

jdalton: Added it.

Issue #90, opened 2017-03-24
Code review notes


jdalton:
I wanted to capture the notes I took during our code review.

## Moved 
- An option to handle proposed syntax (nested import, namespace export) (moved to #100)
- Don't save setter for const exports. (moved to #134)

## Completed
- [x] Check if namespace import is constant and if so the bind isn't needed and it can be const.
- [x] Use arrow functions for module.import and `module.exports`, `module.import`, and `module.importSync`.
- [x] The namespace object should be `Object.create(null)`
- [x] The cache hash - look into mtime + path + version.
- [x] If a module has `import`/`export` add `'use strict'`
- [x] Align setEsModule `__esModule` property non-enumerable
- [x] ~~- Make the uniqueKey a fast property by changing the `:` to a `__` or something similar~~
- [x] ~~Should `addGetters` throw? Yes.~~
- [x] `forEachSetter` (use UNDEFINED lookup)

jdalton:
> Check if namespace import is constant and if so the bind isn't needed and it can be const.

The rules for namespace properties are [here](http://www.ecma-international.org/ecma-262/7.0/#sec-module-namespace-exotic-objects). The property descriptors should be:
```
{ [[Writable]]: true, [[Enumerable]]: true, [[Configurable]]: false }
```
Which means you can create the namespace object and then pass it through [Object.seal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/seal).

jdalton:
No need to have `Object.defineProperty` for all of them since `Object.seal` will take care of it.

benjamn:
`[[Writable]]: true` though?

benjamn:
I thought that to-do item was more to check if the namespace variable itself could be modified to hold a value different from the original object, which is what the `<setter function>.bind(ns = Object.create(null))` protects against.

jdalton:
> [[Writable]]: true though?

That's what the spec says. I've pinged Brian Terlson to see if it's a spec bug.

> I thought that to-do item was more to check if the namespace variable itself could be changed to a value different from the original object,

Ah ya. Imported identifiers are all `const`.

jdalton:
Ok so, [it's still writable in the draft spec too](http://www.ecma-international.org/ecma-262/7.0/#sec-module-namespace-exotic-objects).

jdalton:
I looked up taking on 
> Make the uniqueKey a fast property by changing the : to a __ or something similar

[...]

PR #101, opened 2017-03-28
Use a cache key involving filename and mtime when possible.

### LINKED ISSUE #98: Investigate double compile.
I’ve been working on a PR to use mtime in the cache name (related to #95) and running into test fails that don't happen if I use the `content`.

For example take `test/name.js`.
There are 2 cache reads (both with the same mtime but different contents).

The first time is the original file:
```js
var path = require("path");

export var id = module.id,
  name = path.basename(__filename);

export function foo() {
  return "foo";
}
```

The second time is the modified contents:
```js
(function (exports, require, module, __filename, __dirname) { module.export({id:function(){return id},name:function(){return name},foo:function(){return foo}});var path = require("path");

var id = module.id,
  name = path.basename(__filename);

function foo() {
  return "foo";
}

});
```

The file itself isn't changes which is why the mtime is unchanged but the content passed to `compile` is. I think there’s misdirection going on.

Between the two cache checks it looks like `Module.load` and `Module._compile` are only called once so maybe it’s coming from another place.

benjamn:
The `vm.*` wrappers registered by [`repl/index.js`](https://github.com/benjamn/reify/blob/master/repl/index.js) still just hash the contents, so those cache entries should no longer collide with the ones used by `Module.prototype._compile`.

Part of #90.
Closes #98.

### DIFF
--- a/node/caching-compiler.js
+++ b/node/caching-compiler.js
@@ -17,27 +17,32 @@ const reifyVersion = require("../package.json")
 
 const DEFAULT_CACHE_DIR = ".reify-cache";
 
-exports.compile = (content, filename) => {
-  if (filename === "repl") {
+exports.compile = (content, options) => {
+  options = options || Object.create(null);
+
+  if (options.filename === "repl") {
     // Treat the REPL as if there was no filename.
-    filename = null;
+    options.filename = null;
   }
 
-  const pkgInfo = filename ? getPkgInfo(filename) : fallbackPkgInfo;
+  const pkgInfo = options.filename
+    ? getPkgInfo(options.filename)
+    : fallbackPkgInfo;
+
   if (! pkgInfo) {
     return content;
   }
 
-  return filename
-    ? readWithCacheAndFilename(pkgInfo, content, filename)
-    : readWithCache(pkgInfo, content);
+  return options.filename
+    ? readWithCacheAndFilename(pkgInfo, content, options)
+    : readWithCache(pkgInfo, content, options);
 };
 
-function readWithCacheAndFilename(pkgInfo, content, filename) {
+function readWithCacheAndFilename(pkgInfo, content, options) {
   try {
-    return readWithCache(pkgInfo, content);
+    return readWithCache(pkgInfo, content, options);
   } catch (e) {
-    e.message += ' while processing file: ' + filename;
+    e.message += ' while processing file: ' + options.filename;
     throw e;
   }
 }
@@ -48,10 +53,13 @@ const fallbackPkgInfo = {
   cache: Object.create(null)
 };
 
-function readWithCache(pkgInfo, content) {
+function readWithCache(pkgInfo, content, options) {
   const json = pkgInfo && pkgInfo.json;
   const reify = json && json.reify;
-  const cacheFilename = getCacheFilename(content, reify);
+
+  const cacheFilename =
+    getCacheFilename(options.cacheKey || content, reify);
+
   const absCachePath = typeof pkgInfo.cacheDir === "string" &&
     path.join(pkgInfo.cacheDir, cacheFilename);
 
@@ -103,16 +111,18 @@ function readFileOrNull(filename) {
   }
 }
 
-function getCacheFilename(content, reify) {
-  const hash = createHash("sha1")
-    .update(reifyVersion).update("\0")
-    .update(content).update("\0");
-
-  if (reify) {
-    hash.update(JSON.stringify(reify));
+function getCacheFilename() {
+  const args = [];
+  const argc = arguments.length;
+  for (let i = 0; i < argc; ++i) {
+    args.push(arguments[i]);
   }
 
-  return hash.digest("hex") + ".js";
+  return createHash("sha1")
+    .update(reifyVersion)
+    .update("\0")
+    .update(JSON.stringify(args))
+    .digest("hex") + ".js";
 }
 
 function getPkgInfo(filename) {
--- a/node/caching-compiler.js
+++ b/node/caching-compiler.js
@@ -57,8 +57,11 @@ function compileWithCache(pkgInfo, content, options) {
   const json = pkgInfo && pkgInfo.json;
   const reify = json && json.reify;
 
-  const cacheFilename =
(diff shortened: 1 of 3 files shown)


node/caching-compiler.js
```diff
@@ -103,16 +111,18 @@ function readFileOrNull(filename) {
...
-function getCacheFilename(content, reify) {
-  const hash = createHash("sha1")
-    .update(reifyVersion).update("\0")
-    .update(content).update("\0");
-
-  if (reify) {
-    hash.update(JSON.stringify(reify));
+function getCacheFilename() {
+  const args = [];
+  const argc = arguments.length;
```
jdalton: other places you use `xyzCount` so prolly consistency with that instead of `argc` maybe `argsCount`.

node/caching-compiler.js
```diff
@@ -103,16 +111,18 @@ function readFileOrNull(filename) {
...
+  for (let i = 0; i < argc; ++i) {
+    args.push(arguments[i]);
```
jdalton: Since this is a straight assignment you can do:
```js
const argsCount = arguments.length;
const args = new Array(argsCount)

for (let i = 0; i < argsCount; ++i) {
  args[i] = arguments[i];
}
```

node/compile-hook.js
```diff
@@ -11,10 +12,31 @@ if (! _compile.reified) {
...
+    options.cacheKey = {
+      source: "Module.prototype._compile",
+      filename: filename,
+      mtime: mtime
+    };
+  }
+
+  return options;
+}
+
```
jdalton: I think there's already a `statOrNull` helper somewhere maybe a `node/util.js` would be handy.

node/compile-hook.js
```diff
@@ -11,10 +12,31 @@ if (! _compile.reified) {
...
+function getOptions(filename) {
+  const options = { filename: filename };
+  const mtime = mtimeOrNull(filename);
+
+  if (mtime !== null) {
+    options.cacheKey = {
+      source: "Module.prototype._compile",
+      filename: filename,
+      mtime: mtime
+    };
```
jdalton: JSON stringifying a date will produce a string.
```js
JSON.stringify({ 'a': new Date })
// => '{"a":"2017-03-28T21:36:01.892Z"}'
```
You could have it be a timestamp which I think is a bit better
```js
mtime: mtime.getTime()
```

node/compile-hook.js
```diff
@@ -11,10 +12,31 @@ if (! _compile.reified) {
...
   }).reified = _compile;
 }
+
+function getOptions(filename) {
+  const options = { filename: filename };
+  const mtime = mtimeOrNull(filename);
+
+  if (mtime !== null) {
+    options.cacheKey = {
+      source: "Module.prototype._compile",
```
jdalton: What's the `source` bit do?

node/caching-compiler.js
```diff
@@ -17,27 +17,32 @@ const reifyVersion = require("../package.json")
...
 
-exports.compile = (content, filename) => {
-  if (filename === "repl") {
+exports.compile = (content, options) => {
+  options = options || Object.create(null);
+
+  if (options.filename === "repl") {
     // Treat the REPL as if there was no filename.
-    filename = null;
+    options.filename = null;
```
jdalton: We shouldn't mutate the `options` object. It can be avoided by using `options = Object.assign(Object.create(null), options)` or by referencing the options as one off vars.

node/caching-compiler.js
```diff
@@ -17,27 +17,32 @@ const reifyVersion = require("../package.json")
...
-    ? readWithCacheAndFilename(pkgInfo, content, filename)
-    : readWithCache(pkgInfo, content);
+  return options.filename
+    ? readWithCacheAndFilename(pkgInfo, content, options)
+    : readWithCache(pkgInfo, content, options);
 };
 
-function readWithCacheAndFilename(pkgInfo, content, filename) {
+function readWithCacheAndFilename(pkgInfo, content, options) {
   try {
```
jdalton: When I was working with this for the PR I didn't really dig the names of the methods. I was like `readWithCacheAnd....` then I saw it was also writing the file and compiling. I dig the separation to allow try-catch for just the case where a filename is known but didn't bikeshed a better name yet.

node/caching-compiler.js
```diff
@@ -103,16 +111,18 @@ function readFileOrNull(filename) {
...
+  const argc = arguments.length;
+  for (let i = 0; i < argc; ++i) {
+    args.push(arguments[i]);
   }
 
-  return hash.digest("hex") + ".js";
+  return createHash("sha1")
+    .update(reifyVersion)
+    .update("\0")
+    .update(JSON.stringify(args))
```
jdalton: There are some valid JS `content` values which are not valid inside string literals (like one or two bizarro encodings). That makes me a little worried about using `JSON.stringify` here. It also seems super heavy since we know the arg count is 2 and type detection for one of cacheKey or content is pretty straight forward.

jdalton: Maybe just `JSON.stringify` on `reify` when it exists since when it does it will be a valid JSON object.

node/caching-compiler.js
```diff
@@ -103,16 +111,18 @@ function readFileOrNull(filename) {
...
-    hash.update(JSON.stringify(reify));
+function getCacheFilename() {
+  const args = [];
+  const argc = arguments.length;
+  for (let i = 0; i < argc; ++i) {
+    args.push(arguments[i]);
   }
 
-  return hash.digest("hex") + ".js";
+  return createHash("sha1")
```
jdalton: If it helps with flow you can create the values to add to `createHash("sha1")` as an array and then do
`createHash("sha1").update(values.join("\0")).digest("hex")`

node/compile-hook.js
benjamn: Just a salt to distinguish this call site from any other places where we might call `cachingCompiler.compile`.

node/caching-compiler.js
benjamn: Agreed.

jdalton:
Rock!

PR #102, opened 2017-03-29
Use more ES6 in lib.

jdalton:
This goes through the lib files and modernizes them where safe to do so. It also avoids slower ES5 array built-in use and unneeded slower binds.

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -1,7 +1,6 @@
 "use strict";
 
 const assert = require("assert");
-const undefined = void 0;
 
 // A simplified version of the FastPath AST traversal abstraction used by
 // Recast: https://github.com/benjamn/recast/blob/master/lib/fast-path.js
@@ -118,14 +117,16 @@ FPp.call = function call(callback/*, name1, name2, ... */) {
   const s = this.stack;
   const origLen = s.length;
   let value = s[origLen - 1];
-  const argc = arguments.length;
-  for (let i = 1; i < argc; ++i) {
+  const argCount = arguments.length;
+
+  for (let i = 1; i < argCount; ++i) {
     const name = arguments[i];
     value = value[name];
     s.push(name, value);
   }
   const result = callback(this);
   s.length = origLen;
+
   return result;
 };
 
@@ -137,9 +138,9 @@ FPp.each = function each(callback/*, name1, name2, ... */) {
   const s = this.stack;
   const origLen = s.length;
   const value = s[origLen - 1];
-  const argc = arguments.length;
+  const argCount = arguments.length;
 
-  for (let i = 1; i < argc; ++i) {
+  for (let i = 1; i < argCount; ++i) {
     const name = arguments[i];
     value = value[name];
     s.push(name, value);
@@ -166,9 +167,9 @@ FPp.some = function some(callback/*, name1, name2, ... */) {
   const s = this.stack;
   const origLen = s.length;
   const value = s[origLen - 1];
-  const argc = arguments.length;
+  const argCount = arguments.length;
 
-  for (let i = 1; i < argc; ++i) {
+  for (let i = 1; i < argCount; ++i) {
     const name = arguments[i];
     value = value[name];
     s.push(name, value);
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -1,12 +1,12 @@
 "use strict";
 
-var assert = require("assert");
-var MagicString = require("magic-string/dist/magic-string.cjs.js");
-var utils = require("./utils.js");
-var Visitor = require("./visitor.js");
-var getOption = require("./options.js").get;
+const assert = require("assert");
+const MagicString = require("magic-string/dist/magic-string.cjs.js");
+const utils = require("./utils.js");
+const Visitor = require("./visitor.js");
+const getOption = require("./options.js").get;
 
-var hasOwn = Object.prototype.hasOwnProperty;
+const hasOwn = Object.prototype.hasOwnProperty;
 
 function ImportExportVisitor() {
   Visitor.call(this);
@@ -14,7 +14,7 @@ function ImportExportVisitor() {
 
 module.exports = ImportExportVisitor;
 
-var IEVp = ImportExportVisitor.prototype =
+const IEVp = ImportExportVisitor.prototype =
   Object.create(Visitor.prototype);
 IEVp.constructor = ImportExportVisitor;
 
@@ -44,9 +44,9 @@ IEVp.makeUniqueKey = function () {
 
 IEVp.pad = function (newCode, oldStart, oldEnd) {
   if (this.code) {
-    var oldLines = this.code.slice(oldStart, oldEnd).split("\n");
-    var newLines = newCode.split("\n");
-    var diff = oldLines.length - newLines.length;
+    const oldLines = this.code.slice(oldStart, oldEnd).split("\n");
+    const newLines = newCode.split("\n");
+    let diff = oldLines.length - newLines.length;
     while (diff --> 0) {
       newLines.push("");
     }
@@ -64,7 +64,7 @@ IEVp.overwrite = function (oldStart, oldEnd, newCode, trailing) {
   assert.strictEqual(typeof oldEnd, "number");
   assert.strictEqual(typeof newCode, "string");
 
-  var padded = this.pad(newCode, oldStart, oldEnd);
+  const padded = this.pad(newCode, oldStart, oldEnd);
 
   if (oldStart === oldEnd) {
     if (padded === "") {
@@ -142,7 +142,10 @@ IEVp.getBlockBodyInfo = function (path) {
   // "use strict", which may depend on occurring at the beginning of
   // their enclosing scopes.
   let insertNodeIndex = 0;
-  body.some(function (stmt, i) {
+  const stmtCount = body.length;
+
+  for (let i = 0; i < stmtCount; ++i) {
+    const stmt = body[i];
     if (stmt.type === "ExpressionStatement") {
       const expr = stmt.expression;
       if (expr.type === "StringLiteral" ||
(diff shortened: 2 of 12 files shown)


lib/fast-path.js
```diff
@@ -1,7 +1,6 @@
 "use strict";
 
 const assert = require("assert");
-const undefined = void 0;
```
benjamn: Why remove this?

lib/import-export-visitor.js
```diff
@@ -142,27 +142,32 @@ IEVp.getBlockBodyInfo = function (path) {
   // "use strict", which may depend on occurring at the beginning of
   // their enclosing scopes.
   let insertNodeIndex = 0;
-  body.some(function (stmt, i) {
+  const bodyCount = body.length;
```
benjamn: This should maybe be `stmtCount`, since the `body` array contains statements.

lib/import-export-visitor.js
```diff
@@ -303,15 +318,16 @@ IEVp.finalizeHoisting = function () {
...
   // Just in case we call finalizeHoisting again, don't hoist anything.
   this.bodyInfos.length = 0;
 
-  this.removals.forEach(processRemoval);
+  let removalCount = this.removals.length;
 
-  // Just in case we call finalizeHoisting again, don't remove anything.
-  this.removals.length = 0;
+  while (removalCount--) {
+    processRemoval(this.removals.pop());
```
benjamn: Going in reverse order could be a meaningful semantic change. Also, calling `.pop()` with each iteration seems like a step in the wrong direction, performance-wise. Can we just keep the `forEach` please? The `processRemoval` function is the same every time.

lib/import-export-visitor.js
```diff
@@ -403,7 +423,7 @@ IEVp.visitExportDefaultDeclaration = function (path) {
     // like a comma-separated sequence expression.
     this.overwrite(decl.start, dd.start, exportDefaultPrefix);
 
-    path.call(this.visitWithoutReset, "declaration");
+    path.call((p) => this.visitWithoutReset(p), "declaration");
```
benjamn: `this.visitWithoutReset` is already bound to `this`, so this new arrow function is unnecessary.

lib/import-export-visitor.js
```diff
@@ -593,17 +636,21 @@ function addToSpecifierMap(map, __ported, local) {
 
 function toModuleImport(source, specifierMap, uniqueKey) {
   const parts = ["module.importSync(", source];
-  const importedNames = specifierMap && Object.keys(specifierMap);
+  const importedNames = specifierMap ? Object.keys(specifierMap) : [];
```
benjamn: This is creating an unnecessary empty array.

lib/import-export-visitor.js
```diff
@@ -666,17 +713,19 @@ function safeParam(param, locals) {
 }
 
 function toModuleExport(specifierMap) {
-  const exportedKeys = specifierMap && Object.keys(specifierMap);
+  const exportedKeys = specifierMap ? Object.keys(specifierMap) : [];
```
benjamn: Another unnecessary empty array.

lib/visitor.js
```diff
@@ -6,11 +6,7 @@ const underscoreCode = "_".charCodeAt(0);
 // This Visitor API was inspired by a similar API provided by ast-types:
 // https://github.com/benjamn/ast-types/blob/master/lib/path-visitor.js
 
-function Visitor() {
-  this.visit = this.visit.bind(this);
-  this.visitWithoutReset = this.visitWithoutReset.bind(this);
-  this.visitChildren = this.visitChildren.bind(this);
-}
```
benjamn: Let's keep this binding so that we don't have to keep creating arrow functions every time we call one of these methods.

lib/visitor.js
```diff
@@ -23,7 +19,7 @@ Vp.visit = function (path) {
 Vp.visitWithoutReset = function (path) {
   const value = path.getValue();
   if (Array.isArray(value)) {
-    path.each(this.visitWithoutReset);
+    path.each((p) => this.visitWithoutReset(p));
```
benjamn: Again, no need for an arrow function here (or below).

lib/visitor.js
benjamn: If you really think `p => this.method(p)` is faster than `this.method.bind(this)`, you can create those arrow functions here, but I'd like to see some evidence it matters.

lib/fast-path.js
jdalton: In ES5+ enviros `undefined` is not writable.

lib/import-export-visitor.js
jdalton: cool.

lib/import-export-visitor.js
jdalton: It's type consistency. I may be able to get away with explicit a `null` engines like consistent types.
I ran into perf issues in Firefox due to it and just saw [this](https://twitter.com/bmeurer/status/846951275480711168) yesterday. How often is specifier map provided (what's the common case).

lib/import-export-visitor.js
jdalton: The cost of an empty array is nil compared to an engine invalidating JITed code because of type inconsistencies.

lib/visitor.js
jdalton: Using `bind` is expensive in v8. It was only recently optimized in Node 7. If it can be avoided it should. An arrow function is cheap and this is a parser where performance matters.

benjamn: Can you run the `reify-perf-check` tests with and without this change? I have trouble believing `Function.prototype.bind` is so slow that it should be avoided as a rule of thumb.

jdalton: It's a common v8 gotcha. See https://github.com/Level/levelup/pull/139.

jdalton: I'll give the benchmark a go though.

lib/import-export-visitor.js
jdalton: I was wondering if order mattered. It didn't fail tests. If it's removing things could it have a meaningful change?
Can you give an example? In regards to `pop`, arrays :heart: `pop`, setting the `.length` isn't ness a happy path for engines.

benjamn: I don't think this is a hot spot in the code.

With a top-level parse, this loop happens once per file (because there's only one scope where we hoist anything), and the length of `this.removals` is the number of `import` and `export` declarations we're replacing. Setting `this.removals.length = 0` once per file is never going to be a performance bottleneck, and if it was, we could just as easily not modify the array at all (no `.pop()`, no `.length = 0`, just the loop).

Given that, I would prefer shorter code (`this.removals.forEach(processRemoval)`).

lib/visitor.js
jdalton: It looks like it's `~50ms` win to replace. I can actually replace it in the `Visitor` constructor though. So no need to use arrow functions with every `call(...)`.

jdalton: In Node 4.0.0.

lib/import-export-visitor.js
jdalton: I'm not a fan of micro-optimizations but with everything there's context. In general use I'd say `Array#forEach`, `bind`, and all the built-ins are great. But for something specialized like a game engine or a parser micro-opts can have a cumulative impact. The things I'm suggesting aren't as brittle as other micro-opts. These are things that have held up for 12+ years so less likely to change from release to release of the engine.

jdalton: btw I didn't see your reply about once per file. I ok with omitting it now to get the greater win of more ES6 but it'll be something I think I'll run into again with, know wait a minute is built-in use good here? (as I forget the context of the call).

lib/visitor.js
benjamn: Ok great!

lib/visitor.js
```diff
@@ -7,9 +7,23 @@ const underscoreCode = "_".charCodeAt(0);
...
-  this.visitChildren = this.visitChildren.bind(this);
+  const that = this;
+  const visit = that.visit;
+  const visitWithoutReset =  that.visitWithoutReset;
+  const visitChildren = that.visitChildren;
+
+  // Avoid slower `Function#bind` in Node < 7.
+  this.visit = function () {
+    visit.apply(that, arguments);
+  };
```
jdalton: Is the arg signature of `visit` potentially infinite? or does it max out at like 3 args?

lib/import-export-visitor.js
jdalton: Here's a bit for reference https://github.com/lodash/lodash/commit/861aaf06c62a84e9c3ba952d9fa8d2d531ac3643. I can get away with `null` or `undefined` here if an empty array is too gross.

jdalton: `null` is the non object object and `undefined` is the non value value.

jdalton:
All changes made.

lib/import-export-visitor.js
```diff
@@ -593,17 +635,21 @@ function addToSpecifierMap(map, __ported, local) {
 
 function toModuleImport(source, specifierMap, uniqueKey) {
   const parts = ["module.importSync(", source];
-  const importedNames = specifierMap && Object.keys(specifierMap);
+  const importedNames = specifierMap ? Object.keys(specifierMap) : null;
+  const nameCount = importedNames ? importedNames.length : 0;
```
benjamn: Cool.

lib/visitor.js
```diff
@@ -7,9 +7,23 @@ const underscoreCode = "_".charCodeAt(0);
...
+
+  this.visitWithoutReset = function (path) {
+    visitWithoutReset.call(that, path);
+  };
+
+  this.visitChildren = function (path) {
+    visitChildren.call(that, path);
+  };
```
benjamn: This works for me!

PR #103, opened 2017-03-30
Fix some const and function name assignment bugs.

jdalton:
I decided to try to use babel on reify's bundle (experimenting with smth) and noticed build fails related to assignments to const values.

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -13,7 +13,7 @@ const FPp = FastPath.prototype;
 module.exports = FastPath;
 
 // Static convenience function for coercing a value to a FastPath.
-FastPath.from = function(obj) {
+FastPath.from = function (obj) {
   return obj instanceof FastPath
     // Return a defensive copy of any existing FastPath instances.
     ? obj.copy()
@@ -21,15 +21,15 @@ FastPath.from = function(obj) {
     : new FastPath(obj);
 };
 
-FPp.copy = function copy() {
-  const copy = Object.create(FastPath.prototype);
-  copy.stack = this.stack.slice(0);
-  return copy;
+FPp.copy = function () {
+  const cp = Object.create(FastPath.prototype);
+  cp.stack = this.stack.slice(0);
+  return cp;
 };
 
 // The name of the current property is always the penultimate element of
 // this.stack, and always a String.
-FPp.getName = function getName() {
+FPp.getName = function () {
   const s = this.stack;
   const len = s.length;
   if (len > 1) {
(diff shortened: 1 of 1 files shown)


lib/fast-path.js
```diff
@@ -13,23 +13,23 @@ const FPp = FastPath.prototype;
...
 };
 
-FPp.copy = function copy() {
-  const copy = Object.create(FastPath.prototype);
-  copy.stack = this.stack.slice(0);
-  return copy;
+FPp.copy = function () {
+  const cp = Object.create(FastPath.prototype);
+  cp.stack = this.stack.slice(0);
+  return cp;
```
jdalton: Babel bugged out on the using a const of `copy` since the named function expression was also `copy`.

lib/fast-path.js
```diff
@@ -134,11 +134,11 @@ FPp.call = function call(callback/*, name1, name2, ... */) {
 // accessing this.getValue()[name1][name2]... should be array-like. The
 // callback will be called with a reference to this path object for each
 // element of the array.
-FPp.each = function each(callback/*, name1, name2, ... */) {
+FPp.each = function (callback/*, name1, name2, ... */) {
   const s = this.stack;
   const origLen = s.length;
-  const value = s[origLen - 1];
   const argCount = arguments.length;
+  let value = s[origLen - 1];
```
jdalton: `const value` should be `let value`.

lib/fast-path.js
```diff
@@ -163,11 +163,11 @@ FPp.each = function each(callback/*, name1, name2, ... */) {
 // accessing this.getValue()[name1][name2]... should be array-like. The
 // callback will be called with a reference to this path object for each
 // element of the array.
-FPp.some = function some(callback/*, name1, name2, ... */) {
+FPp.some = function (callback/*, name1, name2, ... */) {
   const s = this.stack;
   const origLen = s.length;
-  const value = s[origLen - 1];
   const argCount = arguments.length;
+  let value = s[origLen - 1];
```
jdalton: Same here.

jdalton:
I removed the names from the named function expressions to match code style of reify.

lib/fast-path.js
```diff
@@ -13,23 +13,23 @@ const FPp = FastPath.prototype;
 module.exports = FastPath;
 
 // Static convenience function for coercing a value to a FastPath.
-FastPath.from = function(obj) {
+FastPath.from = function (obj) {
```
benjamn: I appreciate your attention to detail here!

lib/fast-path.js
benjamn: Could call it `newPath` for greater clarity.

jdalton: It could but in another PR after this I'm removing some dead code (which `FFp.copy` will become) so what it's called won't matter too much in the next PR.

PR #104, opened 2017-03-30
Remove dead code.

jdalton:
This PR removes some dead code.

### DIFF
--- a/lib/fast-path.js
+++ b/lib/fast-path.js
@@ -14,17 +14,7 @@ module.exports = FastPath;
 
 // Static convenience function for coercing a value to a FastPath.
 FastPath.from = function (obj) {
-  return obj instanceof FastPath
-    // Return a defensive copy of any existing FastPath instances.
-    ? obj.copy()
-    // Otherwise use obj as the value of the new FastPath instance.
-    : new FastPath(obj);
-};
-
-FPp.copy = function () {
-  const cp = Object.create(FastPath.prototype);
-  cp.stack = this.stack.slice(0);
-  return cp;
+  return new FastPath(obj);
 };
 
 // The name of the current property is always the penultimate element of
@@ -56,12 +46,6 @@ FPp.getContainer = function () {
   return null;
 };
 
-FPp.valueIsDuplicate = function () {
-  const s = this.stack;
-  const valueIndex = s.length - 1;
-  return s.lastIndexOf(s[valueIndex], valueIndex - 1) >= 0;
-};
-
 function getNodeHelper(path, count) {
   const s = path.stack;
 
@@ -80,7 +64,7 @@ function getNodeHelper(path, count) {
 // maintaining a complete list of type names is one of the reasons we're
 // using the FastPath/Visitor abstraction in the first place.
 function isNodeLike(value) {
-  return value &&
+  return value != null &&
     typeof value === "object" &&
     ! Array.isArray(value) &&
     isCapitalized(value.type);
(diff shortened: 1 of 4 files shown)


lib/fast-path.js
```diff
@@ -80,7 +64,7 @@ function getNodeHelper(path, count) {
 // maintaining a complete list of type names is one of the reasons we're
 // using the FastPath/Visitor abstraction in the first place.
 function isNodeLike(value) {
-  return value &&
+  return value != null &&
```
jdalton: This ensures a consistent boolean type is returned.

lib/fast-path.js
```diff
@@ -99,12 +83,12 @@ FPp.valueIsNode = function () {
 };
 
 FPp.getNode = function (count) {
-  if (count === undefined) count = 0;
+  count = typeof count === "undefined" ? 0 : count;
```
jdalton: I noticed in other places you use `typeof` for `"undefined"` checks so this is just for consistency. Though it could still be an `if` block if you want.

lib/fast-path.js
```diff
@@ -147,52 +131,16 @@ FPp.each = function (callback/*, name1, name2, ... */) {
...
 
   for (let i = 0; i < value.length; ++i) {
-    if (i in value) {
-      s.push(i, value[i]);
-      // If the callback needs to know the value of i, call
-      // path.getName(), assuming path is the parameter name.
-      callback(this);
-      s.length -= 2;
-    }
+    s.push(i, value[i]);
```
jdalton: Since we aren't dealing with sparse arrays the `i in value` check isn't needed.

lib/runtime.js
```diff
@@ -70,18 +69,17 @@ exports.enable = function (Module) {
   // inside loops. The compiler generates these keys automatically (and
   // deterministically) when compiling nested import declarations.
   Mp.importSync = function (id, setters, key) {
-    var module = this;
-    setESModule(module);
+    setESModule(this);
```
jdalton: Since there are no closures now and we've tripped up over the local `module` var before I went ahead and made it just `this` so it's more clear when we're looking at it that it's not the other `module`.

lib/runtime.js
```diff
@@ -100,18 +98,17 @@ exports.enable = function (Module) {
...
   // and the values are functions that return local values.
   Mp.export = function (getters) {
-    var module = this;
-    setESModule(module);
+    setESModule(this);
 
-    if (utils.isPlainObject(getters)) {
-      Entry.getOrCreate(module.id).addGetters(getters);
+    if (getters && typeof getters === "object") {
+      Entry.getOrCreate(this.id).addGetters(getters);
```
jdalton: Instead of the heavier `utils.isPlainObject` I went with the lighter object check used everywhere else in the code.

lib/fast-path.js
```diff
@@ -14,17 +14,7 @@ module.exports = FastPath;
...
-    ? obj.copy()
-    // Otherwise use obj as the value of the new FastPath instance.
-    : new FastPath(obj);
-};
-
-FPp.copy = function () {
-  const cp = Object.create(FastPath.prototype);
-  cp.stack = this.stack.slice(0);
-  return cp;
+  return new FastPath(obj);
```
jdalton: Since `FastPath.from` is under our control we never call it with a `FastPath` instance so it can be simplified and `FPp.copy` can be removed. As with all this if we have a need for it later on it can always be added.

PR #106, opened 2017-03-30
Let first getter win if one already exists.

jdalton:
This addresses a line item of #90.

### DIFF
--- a/lib/entry.js
+++ b/lib/entry.js
@@ -35,8 +35,6 @@ function makeUniqueKey() {
 }
 
 Ep.addSetters = function (parent, setters, key) {
-  var entry = this;
-
   if (typeof key === "undefined") {
     // If no key was provided, make a new unique key that won't collide
     // with any other keys.
@@ -57,15 +55,14 @@ Ep.addSetters = function (parent, setters, key) {
         // Ignore any requests for the exports.__esModule property."
         name !== "__esModule") {
       setter.parent = parent;
-      (entry.setters[name] =
-       entry.setters[name] || Object.create(null)
+      (this.setters[name] =
+       this.setters[name] || Object.create(null)
       )[key] = setter;
     }
   }
 };
 
 Ep.addGetters = function (getters) {
-  var entry = this;
   var names = Object.keys(getters);
   var nameCount = names.length;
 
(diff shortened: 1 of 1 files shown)


lib/entry.js
```diff
@@ -75,8 +72,10 @@ Ep.addGetters = function (getters) {
     if (typeof getter === "function" &&
         // Ignore any requests for the exports.__esModule property."
         name !== "__esModule") {
-      // Should this throw if hasOwn.call(this.getters, name)?
-      entry.getters[name] = getter;
+      if (hasOwn.call(this.getters, name)) {
+        throw new Error("A getter for " + name + " already exists");
```
jdalton: The message can be bikeshedded. I just put something in it.

jdalton:
A follow up to this would be a unit test to ensure it throws. I'm not super up on tests here though so it may be easier for @benjamn to do the follow up test.

lib/entry.js
jdalton: Any preference on the error message?

benjamn: Let's make it `Identifier 'foo' exported more than once`?

lib/entry.js
```diff
@@ -75,8 +72,10 @@ Ep.addGetters = function (getters) {
     if (typeof getter === "function" &&
         // Ignore any requests for the exports.__esModule property."
         name !== "__esModule") {
-      // Should this throw if hasOwn.call(this.getters, name)?
-      entry.getters[name] = getter;
+      if (hasOwn.call(this.getters, name)) {
+        throw new SyntaxError("Identifier '" + name + "' exported more than once");
+      }
```
jdalton: Updated!

benjamn: If we could have this error thrown only during tests, and otherwise let the first getter win, I think that would be better than always throwing, since this is a runtime exception enforcing a compile-time guarantee.

lib/entry.js
```diff
@@ -74,9 +71,10 @@ Ep.addGetters = function (getters) {
     var getter = getters[name];
     if (typeof getter === "function" &&
         // Ignore any requests for the exports.__esModule property."
-        name !== "__esModule") {
+        name !== "__esModule" &&
+        ! hasOwn.call(this.getters, name)) {
       // Should this throw if hasOwn.call(this.getters, name)?
-      entry.getters[name] = getter;
+      this.getters[name] = getter;
     }
```
jdalton: punted on erroring for now, opting to at least let the first getter win.

PR #107, opened 2017-03-30
Defer loading parsers/default.

jdalton:
I noticed the acorn parser was loading even though I was only specifying the top-level.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -8,8 +8,6 @@ const shebangRegExp = /^#!.*/;
 const importExportRegExp = /\b(?:im|ex)port\b/;
 const getOption = require("./options.js").get;
 
-exports.parse = require("./parsers/default.js").parse;
-
 function compile(code, options) {
   const parse = getOption(options, "parse");
   code = code.replace(shebangRegExp, "");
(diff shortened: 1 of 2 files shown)


lib/compiler.js
```diff
@@ -8,8 +8,6 @@ const shebangRegExp = /^#!.*/;
 const importExportRegExp = /\b(?:im|ex)port\b/;
 const getOption = require("./options.js").get;
 
-exports.parse = require("./parsers/default.js").parse;
-
```
jdalton: It looks like `compiler.parse` isn't being used. Since it's configurable in `compile` I opted for the easy route of simply removing the export.

lib/options.js
```diff
@@ -3,9 +3,11 @@
 const hasOwn = Object.prototype.hasOwnProperty;
 
 const defaultCompileOptions = {
+  ast: false,
   generateLetDeclarations: false,
-  parse: require("./parsers/default.js").parse,
-  ast: false
+  get parse() {
+    return require("./parsers/default.js").parse;
```
jdalton: a `require` call is cached so this isn't a gross perf thing.

PR #108, opened 2017-04-02
Add .mjs support.

jdalton:
This PR adds `.mjs` support. It will need a unit test. If you add one I can mimic for others.

### DIFF
--- a/node/caching-compiler.js
+++ b/node/caching-compiler.js
@@ -79,8 +79,12 @@ function compileWithCache(pkgInfo, content, options) {
     }
   }
 
+  const filename = options.filename;
+  const force = !!filename && path.extname(filename) === ".mjs";
+
   const compileOptions = {
-    ast: false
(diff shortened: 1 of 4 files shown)


node/caching-compiler.js
```diff
@@ -246,10 +250,7 @@ function readPkgInfo(dir) {
           for (let i = 0; i < filesCount; ++i) {
             // Later we'll change the value to the actual contents of the
             // file, but for now we merely register that it exists.
-            const file = cacheFiles[i];
-            if (/\.js$/.test(file)) {
-              pkgInfo.cache[file] = true;
-            }
```
benjamn: I think this is okay because we write `.js` cache files even if the original file was `.mjs`.

jdalton: Ya, the check, `/\.js$/`, didn't seem to be needed as all files were `.js` anyways.

PR #110, opened 2017-04-04
Use arrow functions for generated `module.importSync` use.

jdalton:
This PR tackled another item on #90
> Use arrow functions for module.import and `module.exports`, `module.import`, and `module.importSync`.

### DIFF
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -291,7 +291,7 @@ IEVp.finalizeHoisting = function () {
       parts.push(bodyInfo.hoistedImportsString);
     }
 
-    if (parts.length > 0) {
+    if (parts.length) {
       const codeToInsert = parts.join("");
 
       if (this.magicString) {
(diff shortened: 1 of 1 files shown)


lib/import-export-visitor.js
```diff
@@ -652,48 +682,42 @@ function toModuleImport(source, specifierMap, uniqueKey) {
...
-      local = "this";
+      const local = locals[0];
 
+      if (local.includes(".")) {
+        parts.unshift(
+          local,
+          "=Object.create(null);"
+        );
+      }
       // When the imported name is "*", the setter function may be called
```
jdalton: I'm not super excited about `local.includes(".")`. Can you think of a better check?

lib/import-export-visitor.js
```diff
@@ -355,21 +355,51 @@ function processRemoval(removal) {
...
+          identifiers[i],
+          isLast ? ";" : ","
+        );
+      }
+    }
+
+    const namespaceCount = namespaces.length;
+    if (namespaceCount) {
+      const lastIndex = namespaceCount - 1;
+      parts.push("const ");
```
benjamn: I think this should be `this.generateLetDeclarations ? "const " : "var "` for the REPL (and older versions of Node, though that's less important).

lib/import-export-visitor.js
```diff
@@ -390,7 +420,7 @@ IEVp.visitExportAllDeclaration = function (path) {
     this.pad("module.importSync(", decl.start, decl.source.start),
     this._getSourceString(decl),
     this.pad(
-      ",{'*':function(v,k){exports[k]=v;}}," +
+      ",{'*':(v,k) => {exports[k]=v;}}," +
```
benjamn: If we're trying to save characters, let's drop the whitespace here.

lib/import-export-visitor.js
```diff
@@ -652,48 +682,39 @@ function toModuleImport(source, specifierMap, uniqueKey) {
...
+        parts.unshift(`${local}=Object.create(null);`);
+      }
       // When the imported name is "*", the setter function may be called
       // multiple times, and receives an additional parameter specifying
       // the name of the property to be set.
       const nameParam = safeParam("n", [local, valueParam]);
 
       parts.push(
-        ",", nameParam, "){",
+        ",", nameParam, ") => {",
```
benjamn: whitespace

lib/import-export-visitor.js
```diff
@@ -652,48 +682,39 @@ function toModuleImport(source, specifierMap, uniqueKey) {
...
         // The local variable should have been initialized as an empty
         // object when the variable was declared.
         local, "[", nameParam, "]=", valueParam
       );
 
     } else {
       // Multiple local variables become a compound assignment.
-      parts.push("){", locals.join("="), "=", valueParam);
+      parts.push(") => {", locals.join("="), "=", valueParam);
```
benjamn: whitespace

lib/import-export-visitor.js
```diff
@@ -731,9 +752,9 @@ function toModuleExport(specifierMap) {
 
     parts.push(
       exported,
-      ":function(){return ",
+      ":() => ",
```
benjamn: Maybe just `":_=>"`?

lib/import-export-visitor.js
```diff
@@ -652,48 +682,39 @@ function toModuleImport(source, specifierMap, uniqueKey) {
...
-
-      // We must be handling either an ImportNamespaceSpecifier or an
-      // ExportNamespaceSpecifier. We initialize the target object as the
-      // argument to a .bind method call on the setter function, so that
-      // we can refer to it as `this` inside the setter function.
-      bind = ".bind(" + local + "=Object.create(null))";
-      local = "this";
+      const local = locals[0];
 
+      if (local.includes(".")) {
```
benjamn: This might be easier to read if it was `local.startsWith("exports.")`, since that's what we care about.

lib/import-export-visitor.js
benjamn: Ah, that was a bad idea, since `_` could be the thing we're trying to export!

lib/import-export-visitor.js
jdalton: Ok! I wasn't sure about that.

lib/import-export-visitor.js
jdalton: Cool. If there's a way to know this without sniffing the string I'm all for it.

lib/import-export-visitor.js
jdalton: Sure! I'd like to eventually like a repl switch since the repl is not the primary use case it shouldn't dictate the default. Since we support Node 4+ `const` isn't a problem once ESM modules get `"use strict"` added to them (it's a todo).

jdalton:
Updated.

PR #112, opened 2017-04-07
Update tests with more ES6.

jdalton:
This PR adds more ES6.

It'll be interesting to see how more ES6 is handled for `let` and `const` in Node 4 without the implicit `"use strict"` being added yet.

### DIFF
--- a/test/all-files.js
+++ b/test/all-files.js
@@ -10,17 +10,17 @@ import {
   statSync,
 } from "fs";
 
-export var files = Object.create(null);
+export const files = Object.create(null);
 
 function walk(dir) {
-  readdirSync(dir).forEach(function (item) {
-    var absPath = join(dir, item);
-    var stat = statSync(absPath);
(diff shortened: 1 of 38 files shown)


jdalton:
Ah yep.

benjamn:
I think we should be adding new tests instead of changing existing tests. No reason to discard/distort the information we were getting from old tests (and make it harder to run the tests in older versions of Node) just to make the tests look more modern.

jdalton:
Reify only supports Node 4+. At the moment Reify lacks support for ES6 `let` and `const` use in Node 4 because it doesn't add the "use strict" directive *(which it should anyways with ESM to get the right behavior in other areas)*.

jdalton:
This PR is highlighting a pretty critical flaw that will prevent it from being used in a wider audience.

jdalton:
🎉 🌮 🎉

PR #114, opened 2017-04-07
Use ES6 classes.

jdalton:
Use more ES6 and ES6 classes.

### DIFF
--- a/lib/assignment-visitor.js
+++ b/lib/assignment-visitor.js
@@ -1,84 +1,75 @@
 "use strict";
 
 const utils = require("./utils.js");
-const hasOwn = Object.prototype.hasOwnProperty;
 const Visitor = require("./visitor.js");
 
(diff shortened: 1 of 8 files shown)


lib/assignment-visitor.js
```diff
@@ -1,84 +1,79 @@
...
-
-module.exports = AssignmentVisitor;
-
-const AVp = AssignmentVisitor.prototype =
-  Object.create(Visitor.prototype);
-AVp.constructor = AssignmentVisitor;
+module.exports = class AssignmentVisitor extends Visitor {
+  constructor() {
+    super()
+  }
```
benjamn: This constructor shouldn't be necessary, since `Visitor.prototype.constructor` will get called directly without it.

jdalton: Ohhh snao! Nice!

PR #115, opened 2017-04-07
Use a simplified magic-string.

jdalton:
This is a start of a simplified magic-string.

### DIFF
--- a/lib/magic-string.js
+++ b/lib/magic-string.js
@@ -1,0 +1,142 @@
+"use strict";
+
+// A simplified version of magic-string.
+// Copyright Rich Harris. Released under MIT license:
+// https://github.com/Rich-Harris/magic-string
+
+class Chunk {
+  constructor(start, end, content) {
+    this.start = start;
+    this.end = end;
+    this.original = content;
+    this.intro = "";
+    this.outro = "";
+    this.content = content;
+    this.next = null;
+  }
+
+  appendLeft(content) {
+    this.outro += content;
+  }
+
+  contains(index) {
+    return this.start < index && index < this.end;
+  }
+
+  edit(content) {
+    this.content = content;
+    this.intro = "";
+    this.outro = "";
+    return this;
+  }
+
+  prependRight(content) {
+    this.intro = content + this.intro;
+  }
+
+  split(index) {
+    const sliceIndex = index - this.start;
+    const originalBefore = this.original.slice(0, sliceIndex);
+    const originalAfter = this.original.slice(sliceIndex);
+
+    const newChunk = new Chunk(index, this.end, originalAfter);
+    newChunk.outro = this.outro;
+
+    this.original = originalBefore;
+    this.outro = "";
+    this.end = index;
+    this.content = originalBefore;
+
+    newChunk.next = this.next;
+    this.next = newChunk;
+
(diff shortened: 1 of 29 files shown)


lib/magic-string.js
```diff
@@ -0,0 +1,142 @@
...
+    this.start = start;
+    this.end = end;
+    this.original = content;
+    this.intro = "";
+    this.outro = "";
+    this.content = content;
+    this.next = null;
+  }
+
+  appendLeft(content) {
```
jdalton: for classes do you prefer `appendLeft (` or `appendLeft(` (space, no space)

lib/magic-string.js
```diff
@@ -0,0 +1,142 @@
...
+
+    while (true) {
+      if (chunk.contains(index)) {
+        this._splitChunk(chunk, index);
+        return;
+      }
+      chunk = searchForward ?
+        this.byStart[chunk.end] :
+        this.byEnd[chunk.start];
+    }
```
jdalton: What's your ternary style preference?

lib/magic-string.js
benjamn: No space after a function or method name. I only put a space before the `(` if it's an anonymous function.

lib/magic-string.js
benjamn:
```js
chunk = searchForward
  ? this.byStart[chunk.end]
  : this.byEnd[chunk.start];
```
please.

lib/magic-string.js
```diff
@@ -0,0 +1,142 @@
...
+    this.byStart[index] = newChunk;
+    this.byEnd[newChunk.end] = newChunk;
+    this.lastSearchedChunk = chunk;
+  }
+
+  toString() {
+    let str = this.intro;
+    let chunk = this.firstChunk;
+    while (chunk) {
+      str += chunk.toString();
```
benjamn: Maybe build up an array of chunks and then `.join("")` at the end?

jdalton:
Ya, I created this build by taking the original magic-string.cjs.js and removed stuff, checked if unit tests failed, then removed more. Having expected output tests though is a good thing. I don't think we need to compare to magic-string proper but have a baseline (which could be this) of what we think is OK so it can catch bugs that may slip in. (I do this with babel-plugin-lodash.

jdalton:
If it helps we can add tests based on the magic-string package output (before this PR) then just rebase and see if they pass.

lib/magic-string.js
jdalton: That optimization has gone out of vogue (no longer needed in modern engines).
The loop as is, is the code magic-string uses.

jdalton: Engines will defer flattening a string until it's necessary. Using `Array#join` forces an immediate flattening.

benjamn: Interesting!

jdalton:
Added a bunch of tests using `"magic-string/dist/magic-string.cjs.js"` results as the expected output.

benjamn:
Awesome! Any feelings about whether we should bump the version to `0.6.11` or `0.7.0`?

jdalton:
This prolly fits a `0.7.0`.

PR #118, opened 2017-04-10
Preserve crlf line endings of code.

jdalton:
I noticed that padding didn't account for crlf. This patch addresses that for those folks on Windows boxes.

### DIFF
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -6,6 +6,7 @@ const MagicString = require("./magic-string.js");
 const utils = require("./utils.js");
 const Visitor = require("./visitor.js");
 
+const carrageReturnCode = "\r".charCodeAt(0);
 const exportDefaultPrefix = "module.export('default',exports.default=(";
 const exportDefaultSuffix = "));";
 const hasOwn = Object.prototype.hasOwnProperty;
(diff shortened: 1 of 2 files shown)


lib/import-export-visitor.js
```diff
@@ -276,11 +277,15 @@ module.exports = class ImportExportVisitor extends Visitor {
     if (this.code) {
       const oldLines = this.code.slice(oldStart, oldEnd).split("\n");
       const newLines = newCode.split("\n");
+      const line = oldLines[0];
+      const endCode = line.charCodeAt(line.length - 1);
+      const lineEnd = endCode === carrageReturnCode ? "\r\n" : "\n";
```
benjamn: The goal of this code is basically to blank out the lines where the `import` or `export` declarations originally appeared, so really what we should be doing is just removing everything from those lines except the line ending. If we did that right, we could just preserve the original line endings, instead of using `oldLines[0]` to detect whether all the lines should end with `\r\n` or `\n`.

jdalton: I wonder if that would be expensive that just detecting the newline?

jdalton: I'll look into doing a match for `/\r?\n/`.

jdalton:
Ah wrong merge.

benjamn:
That's fine, I don't think the comment I made really matters.

jdalton:
Moved to #122.

PR #121, opened 2017-04-10
Fix a perf regression in Node v4.

jdalton:
I noticed that the transpile time had spiked in Node v4 by 500+ms for the benchmark.
After investigation it looks like generated arrow function use in the runtime caused the slow down.
I'll check back in after we add `"use strict"` to see if anything changes.

### DIFF
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -410,7 +410,7 @@ module.exports = class ImportExportVisitor extends Visitor {
       this.pad("module.importSync(", decl.start, decl.source.start),
       this._getSourceString(decl),
       this.pad(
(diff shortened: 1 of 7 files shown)


lib/import-export-visitor.js
```diff
@@ -738,9 +740,9 @@ function toModuleExport(specifierMap) {
 
     parts.push(
       exported,
-      ":()=>",
+      ":function(){return ",
```
benjamn: Consistency is good, but I suspect avoiding arrow functions for `module.export` is less important than for `module.importSync`.

jdalton: It looks like I can get away with not doing it for `module.export`. Do you want to split it?

PR #123, opened 2017-04-11
Add "use strict" to any modules containing import or export declarations

benjamn:
Addresses an item from #90.

Should help with Node 4 tests: https://github.com/benjamn/reify/pull/112#issuecomment-292797463

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -21,13 +21,17 @@ exports.compile = function (code, options) {
 
   const result = { ast: null, code };
 
-  if (! getOption(options, "force") &&  ! importExportRegExp.test(code)) {
+  if (! getOption(options, "force") &&
+      ! importExportRegExp.test(code)) {
+    // Let the caller know the result is no different from the input.
+    options.identical = true;
 
     if (getOption(options, "ast")) {
       // To take full advantage of identity checking, you should probably
       // pass the { ast: false } option to the compile function.
       result.ast = parse(code);
     }
+
     return result;
   }
 
@@ -38,13 +42,19 @@ exports.compile = function (code, options) {
 
   const magicString = importExportVisitor.magicString;
 
-  assignmentVisitor.visit(rootPath, {
-    exportedLocalNames: importExportVisitor.exportedLocalNames,
-    magicString: magicString,
-    modifyAST: importExportVisitor.modifyAST
-  });
+  if (importExportVisitor.madeChanges) {
+    assignmentVisitor.visit(rootPath, {
+      exportedLocalNames: importExportVisitor.exportedLocalNames,
+      magicString: magicString,
+      modifyAST: importExportVisitor.modifyAST
+    });
 
-  importExportVisitor.finalizeHoisting();
+    importExportVisitor.finalizeHoisting();
+
+  } else {
+    // Let the caller know the result is no different from the input.
+    options.identical = true;
+  }
 
   if (importExportVisitor.modifyAST) {
     // If the importExportVisitor modified the AST because options.ast was
@@ -70,12 +80,18 @@ exports.transform = function (ast, options) {
     importExportOptions
   );
 
-  assignmentVisitor.visit(rootPath, {
-    exportedLocalNames: importExportVisitor.exportedLocalNames,
-    modifyAST: true
-  });
+  if (importExportVisitor.madeChanges) {
+    assignmentVisitor.visit(rootPath, {
+      exportedLocalNames: importExportVisitor.exportedLocalNames,
+      modifyAST: true
+    });
 
-  importExportVisitor.finalizeHoisting();
+    importExportVisitor.finalizeHoisting();
+
+  } else {
+    // Let the caller know the result is no different from the input.
+    options.identical = true;
+  }
 
   return ast;
 };
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -19,12 +19,16 @@ exports.compile = function (code, options) {
   const parse = getOption(options, "parse");
   code = code.replace(shebangRegExp, "");
 
-  const result = { ast: null, code };
+  const result = {
+    code,
+    ast: null,
+    identical: false
+  };
 
   if (! getOption(options, "force") &&
       ! importExportRegExp.test(code)) {
     // Let the caller know the result is no different from the input.
-    options.identical = true;
+    result.identical = true;
 
     if (getOption(options, "ast")) {
       // To take full advantage of identity checking, you should probably
@@ -53,7 +57,7 @@ exports.compile = function (code, options) {
 
   } else {
     // Let the caller know the result is no different from the input.
-    options.identical = true;
+    result.identical = true;
   }
 
   if (importExportVisitor.modifyAST) {
@@ -80,6 +84,8 @@ exports.transform = function (ast, options) {
     importExportOptions
   );
 
+  const result = { ast, identical: false };
+
   if (importExportVisitor.madeChanges) {
     assignmentVisitor.visit(rootPath, {
       exportedLocalNames: importExportVisitor.exportedLocalNames,
@@ -90,8 +96,8 @@ exports.transform = function (ast, options) {
 
   } else {
     // Let the caller know the result is no different from the input.
-    options.identical = true;
+    result.identical = true;
   }
 
-  return ast;
+  return result;
 };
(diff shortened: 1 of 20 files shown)


lib/import-export-visitor.js
```diff
@@ -347,11 +394,21 @@ module.exports = class ImportExportVisitor extends Visitor {
...
     return this;
   }
 
+  visitProgram(path) {
+    this.visitChildren(path);
+    path.call(
+      firstStmtPath => this.getBlockBodyInfo(firstStmtPath),
+      "body", 0
+    );
+  }
```
jdalton: What's the `visitProgram` bit do?

benjamn: It makes sure `getBlockBodyInfo` gets called for the `Program`, even if there are no `import` or `export` declarations at the top level.

jdalton: What's the importance of that?

test/output/only-nested/expected.js
```diff
@@ -0,0 +1,3 @@
+"use strict";function f() {var a,c;module.importSync("./module",{"a":function(v){a=v},"b":function(v){c=v}},0);
+
+}
```
jdalton: 👍

lib/import-export-visitor.js
```diff
@@ -114,9 +125,14 @@ module.exports = class ImportExportVisitor extends Visitor {
     const directives = parent.directives;
     if (directives) {
       const directiveCount = directives.length;
-
       for (let i = 0; i < directiveCount; ++i) {
-        insertCharIndex = Math.max(directives[i].end, insertCharIndex);
+        const d = directives[i];
```
jdalton: If Babylon represents directives with `.directives` does Acorn do something different? Is there a potential gap?

jdalton: Oh I see for other parsers it does the `"ExpressionStatement"` check above.

benjamn: Yep, Acorn represents directives as string-valued `ExpressionStatement`s, but that difference only matters from the AST perspective. The generated code looks exactly the same. In other words, if `this.modifyAST` can be assumed to be false, there should be no difference in behavior.

lib/import-export-visitor.js
```diff
@@ -140,13 +157,18 @@ module.exports = class ImportExportVisitor extends Visitor {
...
+
     return bodyInfo;
   }
 
   hoistImports(importDeclPath, hoistedCode, childName) {
     this.preserveChild(importDeclPath, childName);
     const bodyInfo = this.getBlockBodyInfo(importDeclPath);
     bodyInfo.hoistedImportsString += hoistedCode;
+    this.madeChanges = true;
     return this;
```
jdalton: I dig the "madeChanges" as a way to lazily detect if you've hit an `import` or `export`. I wonder if the name should be changed *(bikeshedding)* since it only relates to hoisted import/exports.

test/output/only-nested/expected.js
benjamn: This test case wouldn't work without the `visitProgram` method, because we wouldn't have a `bodyInfo` for the top level (only the `f` function scope).

lib/import-export-visitor.js
jdalton: Maybe `isESM` or `importsOrExportsHoisted` I donno. I get it though.

lib/import-export-visitor.js
```diff
@@ -186,6 +210,13 @@ module.exports = class ImportExportVisitor extends Visitor {
       const bodyInfo = this.bodyInfos[i];
       const parts = [];
 
+      // We don't need to add a "use strict" directive unless the compiler
+      // made any changes.
+      if (this.madeChanges &&
+          bodyInfo.needToAddUseStrictDirective) {
+        parts.push('"use strict";');
```
jdalton: the `.bodyInfos` is reify specific metadata right? Is there a better way to differentiate what is legit AST from what is custom metadata. I noticed while refactoring sometimes I had to take an educated guess/leap.

lib/import-export-visitor.js
```diff
@@ -211,21 +242,37 @@ module.exports = class ImportExportVisitor extends Visitor {
...
+
+          const parsedDirectives = ast.directives;
+          const parentDirectives = bodyInfo.parent.directives;
+
+          if (parsedDirectives &&
+              parentDirectives) {
+            parentDirectives.push.apply(
+              parentDirectives,
+              parsedDirectives
+            );
```
jdalton: Does this only apply for Babylon since it requires the `.directives` property. How are non-Babylon handled?

lib/import-export-visitor.js
```diff
@@ -347,11 +394,21 @@ module.exports = class ImportExportVisitor extends Visitor {
       !! getOption(options, "generateLetDeclarations");
     this.modifyAST = !! getOption(options, "ast");
     this.parse = getOption(options, "parse");
+    this.enforceStrictMode = getOption(options, "enforceStrictMode");
```
jdalton: Is it odd that `enforceStrictMode` is `true` by default while others are `false`. Should this be flipped to loose or sloppy (I honestly don't know which is technically more correct).

lib/import-export-visitor.js
```diff
@@ -468,6 +525,8 @@ module.exports = class ImportExportVisitor extends Visitor {
 
         path.replace(this._buildExportDefaultStatement(dd));
       }
+
+      this.madeChanges = true;
```
benjamn: Here's one case where we set `this.madeChanges = true` even though nothing is being hoisted. Specifically, this means we still add `"use strict"` when you have only `export default someExpression` and no other kinds of `import` or `export` declarations.

jdalton: Could the `enforceStrictMode` property be re-used for this (as to not introduce another prop)?

benjamn: That's a good idea, I think.

jdalton: I guess you need one to indicate strict mode should be used and one to overwrite the intent of the parser.

lib/import-export-visitor.js
benjamn: It's a property of the `importExportVisitor`, not the AST. There's a `_bodyInfoByName` property that we put on the parent node [here](https://github.com/benjamn/reify/blob/919b684f34898fd71e2b7d50ee9cd9190dab10f4/lib/import-export-visitor.js#L139-L142), which is differentiated by the leading `_`.

lib/compiler.js
```diff
@@ -38,13 +42,19 @@ exports.compile = function (code, options) {
 
   const magicString = importExportVisitor.magicString;
 
-  assignmentVisitor.visit(rootPath, {
-    exportedLocalNames: importExportVisitor.exportedLocalNames,
-    magicString: magicString,
-    modifyAST: importExportVisitor.modifyAST
-  });
+  if (importExportVisitor.madeChanges) {
```
benjamn: I think this usage lends a bit more support to the `madeChanges` terminology, yeah?

lib/compiler.js
```diff
@@ -21,13 +21,17 @@ exports.compile = function (code, options) {
 
   const result = { ast: null, code };
 
-  if (! getOption(options, "force") &&  ! importExportRegExp.test(code)) {
+  if (! getOption(options, "force") &&
+      ! importExportRegExp.test(code)) {
+    // Let the caller know the result is no different from the input.
+    options.identical = true;
```
jdalton: There should be a better way to signal that there is no difference since modifying user provided options is generally seen as bad news.

jdalton: Maybe a different kind of result like an object with a `response` or smth that can also have other data props associated with it.

lib/compiler.js
jdalton: Ok!

test/output/only-nested/expected.js
jdalton: Ahhh so if we later toggle nested imports then this can be skipped when toggled off?

lib/compiler.js
```diff
@@ -19,12 +19,16 @@ exports.compile = function (code, options) {
   const parse = getOption(options, "parse");
   code = code.replace(shebangRegExp, "");
 
-  const result = { ast: null, code };
+  const result = {
+    code,
+    ast: null,
+    identical: false
+  };
```
jdalton: So much power!

jdalton:
:shipit:

PR #127, opened 2017-04-12
Avoid quotes on generated keys when possible.

jdalton:
This PR avoids quotes on generated keys and settles on a consistent use of double quotes for generated code.

### DIFF
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -7,7 +7,7 @@ const utils = require("./utils.js");
 const Visitor = require("./visitor.js");
 
 const carriageReturnCode = "\r".charCodeAt(0);
-const exportDefaultPrefix = "module.export('default',exports.default=(";
+const exportDefaultPrefix = 'module.export("default",exports.default=(';
 const exportDefaultSuffix = "));";
 const hasOwn = Object.prototype.hasOwnProperty;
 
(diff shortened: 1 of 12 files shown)


jdalton:
This also settles on a consistent use of double quotes for generated code.

lib/import-export-visitor.js
```diff
@@ -717,6 +717,13 @@ function processRemoval(removal) {
   }
 }
 
+function safeKey(key) {
+  if (/^\d+$|^[_$a-zA-Z]\w*$/.test(key)) {
+    return key;
+  }
+  return JSON.stringify(key);
```
jdalton: This may be too safe. I don't think identifiers can be just numbers as they can be with keys.

lib/import-export-visitor.js
```diff
@@ -717,6 +717,13 @@ function processRemoval(removal) {
   }
 }
 
+function safeKey(key) {
+  if (/^\d+$|^[_$a-zA-Z]\w*$/.test(key)) {
```
benjamn: When is the `^\d+$` part important?

jdalton: That's what I was saying. I don't think it is in this case so it can be removed.

jdalton: Updated!

PR #129, opened 2017-04-12
Move transform to its own module.

jdalton:
Moves the transform method into its own module so that it can be more easily excluded from bundles that don't modify the ast.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -1,9 +1,9 @@
 "use strict";
 
-const assert = require("assert");
+const dynRequire = module.require ? module.require.bind(module) : __non_webpack_require__;
+
 const FastPath = require("./fast-path.js");
 const getOption = require("./options.js").get;
-const utils = require("./utils.js");
 
 const IEV = require("./import-export-visitor.js");
 const importExportVisitor = new IEV;
@@ -11,7 +11,6 @@ const importExportVisitor = new IEV;
 const AV = require("./assignment-visitor.js");
 const assignmentVisitor = new AV;
 
-const hasOwn = Object.prototype.hasOwnProperty;
 const shebangRegExp = /^#!.*/;
 const importExportRegExp = /\b(?:im|ex)port\b/;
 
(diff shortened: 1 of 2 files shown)


lib/compiler.js
```diff
@@ -1,103 +1,77 @@
...
-  if (! getOption(options, "force") &&
-      ! importExportRegExp.test(code)) {
-    // Let the caller know the result is no different from the input.
-    result.identical = true;
-
-    if (getOption(options, "ast")) {
-      // To take full advantage of identity checking, you should probably
-      // pass the { ast: false } option to the compile function.
-      result.ast = parse(code);
+module.exports = {
```
benjamn: Let's stick to adding properties to `exports`.

lib/compiler.js
```diff
@@ -1,103 +1,77 @@
...
 
-    importExportVisitor.finalizeHoisting();
+    return result;
+  },
 
-  } else {
-    // Let the caller know the result is no different from the input.
-    result.identical = true;
+  get transform() {
+    return dynRequire("./transform.js");
```
benjamn: This can just be
```js
exports.transform = function () {
  return dynRequire("./transform.js").apply(this, arguments);
};
```

lib/transform.js
```diff
@@ -0,0 +1,40 @@
...
+
+const FastPath = require("./fast-path.js");
+
+const IEV = require("./import-export-visitor.js");
+const importExportVisitor = new IEV;
+
+const AV = require("./assignment-visitor.js");
+const assignmentVisitor = new AV;
+
+module.exports = function (ast, options) {
```
benjamn: This is fine, though.

lib/compiler.js
jdalton: Ok, but fancy getters are fancy 🎩

lib/compiler.js
```diff
@@ -72,32 +71,5 @@ exports.compile = function (code, options) {
...
-
-    importExportVisitor.finalizeHoisting();
-
-  } else {
-    // Let the caller know the result is no different from the input.
-    result.identical = true;
-  }
-
-  return result;
+  return dynRequire("./transform.js")(ast, options);
```
jdalton: I avoided the `.apply` jazz. Let me know if you'd still like it in.

PR #130, opened 2017-04-12
Reduce some nesting and split `preserveChild` out into 2 methods.

jdalton:
This PR reduces some nesting and splits `preserveChild` out into 2 methods.

### DIFF
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -13,24 +13,22 @@ const hasOwn = Object.prototype.hasOwnProperty;
 
 module.exports = class ImportExportVisitor extends Visitor {
   addExportedLocalNames(specifierMap) {
-    if (specifierMap) {
-      const exportedLocalNames = this.exportedLocalNames;
-      const keys = Object.keys(specifierMap);
-      const keyCount = keys.length;
+    const exportedLocalNames = this.exportedLocalNames;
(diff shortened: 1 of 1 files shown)


lib/import-export-visitor.js
```diff
@@ -338,42 +336,42 @@ module.exports = class ImportExportVisitor extends Visitor {
...
-      path.call(this.visitWithoutReset, childName);
+    path.call(this.visitWithoutReset, childName);
+  }
 
-    } else {
-      if (this.code) {
-        this.overwrite(value.start, value.end, "");
-      }
+  preserveLine(path) {
+    const value = path.getValue();
```
jdalton: Is `preserveLine` a good name?

jdalton: Side note:
Is `preserveChild` an official ast-y name? Or should both really be prefixed with a `_`?

benjamn: They could definitely be private methods. I adopted a convention of prefixing actual AST visitor methods with `visit` so that normal methods could have normal names, but I'm open to more conventions.

jdalton: Cool. I'll update them with `_`. Is `_preserveLine` OK? Or do you have a better name in mind?

jdalton: Ok updated.

PR #133, opened 2017-04-13
Use regular string concats.

jdalton:
This PR makes the switch to plain string concats since it's what magic string and string concats are used elsewhere in reify.

### DIFF
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -17,31 +17,28 @@ module.exports = class ImportExportVisitor extends Visitor {
 
     for (let i = 0; i < infoCount; ++i) {
       const bodyInfo = this.bodyInfos[i];
-      const parts = [];
+      let codeToInsert = "";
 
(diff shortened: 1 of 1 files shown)


lib/import-export-visitor.js
```diff
@@ -17,31 +17,28 @@ module.exports = class ImportExportVisitor extends Visitor {
...
       const bodyInfo = this.bodyInfos[i];
-      const parts = [];
+      let codeToInsert = "";
 
       // We don't need to add a "use strict" directive unless the compiler
       // made any changes.
-      if (this.madeChanges &&
-          bodyInfo.needToAddUseStrictDirective) {
-        parts.push('"use strict";');
+      if (this.madeChanges && bodyInfo.needToAddUseStrictDirective) {
```
jdalton: I noticed sometimes you break conditions into multiple lines and someones not.
I usually use an 80 character ruler to help make line break choices.

PR #137, opened 2017-04-18
Another round of cleanup and optimization.

jdalton:
* Move functions out of others in lib/entry and lib/runtime
 * Align `extMjs` var name with convention in node/compile-hook
 * Use `UNDEFINED` and `NAN` objects in place of literals for setter cache in lib/entry
 * Use a symbol as an "__esModule" flag for better runtime performance

### DIFF
--- a/lib/entry.js
+++ b/lib/entry.js
@@ -1,7 +1,11 @@
 "use strict";
 
+var GETTER_ERROR = {};
+var NAN = {};
+var UNDEFINED = {};
+
 var entryMap = Object.create(null);
-var hasOwn = Object.prototype.hasOwnProperty;
+var keySalt = 0;
 
 function Entry(id) {
   // Same as module.id for this module.
@@ -24,38 +28,29 @@ Entry.getOrCreate = function (id) {
   return entryMap[id] = entryMap[id] || new Entry(id);
 };
 
-var keySalt = 0;
-function makeUniqueKey() {
-  return Math.random()
-    .toString(36)
-    // Add an incrementing salt to help track key ordering and also
-    // absolutely guarantee we never return the same key twice.
-    .replace("0.", ++keySalt + ":");
-}
-
 Ep.addSetters = function (parent, setters, key) {
-  if (typeof key === "undefined") {
-    // If no key was provided, make a new unique key that won't collide
-    // with any other keys.
-    key = makeUniqueKey();
-  } else {
-    // If a key was provided, make sure it is distinct from keys provided
-    // by other parent modules.
-    key = parent.id + ":" + key;
-  }
-
   var names = Object.keys(setters);
   var nameCount = names.length;
+
+  if (! nameCount) {
+    return;
+  }
+
+  // If no key is provided, make a unique key. Otherwise, make sure the key is
+  // distinct from keys provided by other parent modules.
+  key = typeof key === "undefined"
+    ? makeUniqueKey()
+    : parent.id + ":" + key;
 
   for (var i = 0; i < nameCount; ++i) {
     var name = names[i];
     var setter = setters[name];
     if (typeof setter === "function" &&
-        // Ignore any requests for the exports.__esModule property."
+        // Ignore any requests for the exports.__esModule property.
         name !== "__esModule") {
       setter.parent = parent;
       (this.setters[name] =
-       this.setters[name] || Object.create(null)
+        this.setters[name] || Object.create(null)
       )[key] = setter;
     }
   }
@@ -69,43 +64,28 @@ Ep.addGetters = function (getters) {
     var name = names[i];
     var getter = getters[name];
     if (typeof getter === "function" &&
-        // Ignore any requests for the exports.__esModule property."
+        // Ignore any requests for the exports.__esModule property.
         name !== "__esModule" &&
-        // Should this throw if hasOwn.call(this.getters, name)?
-        ! hasOwn.call(this.getters, name)) {
+        // Should this throw if this.getters[name] exists?
+        ! this.getters[name]) {
       this.getters[name] = getter;
     }
   }
 };
-
-function runModuleSetters(module) {
-  var entry = entryMap[module.id];
-  if (entry) {
-    entry.runModuleSetters(module);
-  }
-}
 
 Ep.runModuleGetters = function (module) {
   var names = Object.keys(this.getters);
   var nameCount = names.length;
 
   for (var i = 0; i < nameCount; ++i) {
-    this.runGetter(module, names[i]);
-  }
-};
+    var name = names[i];
+    var value = runGetter(this, name);
 
-// Returns true iff the getter updated module.exports with a new value.
-Ep.runGetter = function (module, name) {
-  if (hasOwn.call(this.getters, name)) {
-    try {
-      // Update module.exports[name] with the current value so that CommonJS
-      // require calls remain consistent with module.importSync.
-      return module.exports[name] =
-        this.getters[name].call(module);
-
-    } catch (e) {
-      // If the getter threw an exception, avoid updating module.exports
-      // and return undefined.
+    // If the getter is run without error, update module.exports[name] with
+    // the current value so that CommonJS require calls remain consistent with
+    // module.importSync.
+    if (value !== GETTER_ERROR) {
+      module.exports[name] = value;
     }
   }
 };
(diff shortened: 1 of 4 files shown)


node/compile-hook.js
```diff
@@ -35,9 +35,9 @@ if (! _compile.reified) {
...
 const exts = Module._extensions;
-const _extMjs = exts[".mjs"];
-if (! (_extMjs && _extMjs.reified)) {
+const extMjs = exts[".mjs"];
+if (! (extMjs && extMjs.reified)) {
   (exts[".mjs"] = (module, filename) =>
     exts[".js"](module, filename)
-  ).reified = _extMjs;
+  ).reified = extMjs;
 }
```
jdalton: Here I noticed the convention didn't require an `_`.

lib/runtime.js
```diff
@@ -125,13 +111,32 @@ exports.enable = function (Module) {
     if (name === "default" &&
         ! (exports &&
            typeof exports === "object" &&
-           exports.__esModule &&
+           hasOwn.call(exports, "__esModule") &&
+           exports.__esModule === true &&
            "default" in exports)) {
```
jdalton: I made the check here align a bit more with how we set it.

lib/runtime.js
```diff
@@ -125,13 +111,32 @@ exports.enable = function (Module) {
...
       return exports;
     }
 
-    return exports && exports[name];
+    if (exports == null) {
+      return;
+    }
```
jdalton: I need to make a test here since `exports` could be `0`

lib/entry.js
```diff
@@ -209,4 +139,75 @@ Ep.runModuleSetters = function (module) {
...
+    // Ignore setters asking for module.exports.__esModule.
+    return;
+  }
+
+  if (value !== value) {
+    value = NAN;
+  } else if (typeof value === "undefined") {
+    value = UNDEFINED;
+  }
+
```
jdalton: I implemented the `UNDEFINED` technique from #90 and added one for `NaN` too.

lib/runtime.js
```diff
@@ -1,5 +1,6 @@
 "use strict";
 
+var __esModule = typeof Symbol === "function" ? Symbol.for("__esModule") : "__esModule";
 var hasOwn = Object.prototype.hasOwnProperty;
```
jdalton: The use of a symbol here avoids `Object.defineProperty` use in modern environments like Node 4+ and gives us a 30-40ms speed boost on loading (per the benchmark).

lib/runtime.js
```diff
@@ -125,13 +112,37 @@ exports.enable = function (Module) {
     if (name === "default" &&
         ! (exports &&
            typeof exports === "object" &&
-           exports.__esModule &&
+           hasOwn.call(exports, __esModule) &&
            "default" in exports)) {
```
jdalton: symbols work in `Object#hasOwnProperty` checks out of the box.

lib/runtime.js
```diff
@@ -125,13 +112,37 @@ exports.enable = function (Module) {
...
+    if (typeof __esModule === "string") {
+      Object.defineProperty(exports, __esModule, {
+        value: true,
+        enumerable: false,
+        writable: false,
+        configurable: true
+      });
+    } else {
+      exports[__esModule] = true;
+    }
```
jdalton: Here is the toggle for symbol or string use. If it's a string then a non-enumerable property is set otherwise a symbol (which is non-enumerable in the traditional sense) is used.

lib/entry.js
```diff
@@ -24,38 +26,27 @@ Entry.getOrCreate = function (id) {
...
+  key = typeof key === "undefined"
+    ? makeUniqueKey()
+    : parent.id + ":" + key;
+
   for (var i = 0; i < nameCount; ++i) {
     var name = names[i];
     var setter = setters[name];
-    if (typeof setter === "function" &&
-        // Ignore any requests for the exports.__esModule property."
-        name !== "__esModule") {
```
jdalton: Since both string and symbols are non-enumerable they'll never show up in an `Object.keys` result so the checks aren't needed.

lib/entry.js
```diff
@@ -24,38 +26,29 @@ Entry.getOrCreate = function (id) {
...
+    : parent.id + ":" + key;
+
   for (var i = 0; i < nameCount; ++i) {
     var name = names[i];
     var setter = setters[name];
     if (typeof setter === "function" &&
-        // Ignore any requests for the exports.__esModule property."
+        // Ignore any requests for the exports.__esModule property.
         name !== "__esModule") {
       setter.parent = parent;
```
jdalton: I kept the enumerable "__esModule" property checks for Babel loose mode support.

lib/runtime.js
```diff
@@ -125,13 +112,39 @@ exports.enable = function (Module) {
...
 
   return Module;
 };
+
+function setESModule(module) {
+  var exports = module.exports;
+  if (exports &&
+      typeof exports === "object" &&
+      ! hasOwn.call(exports, "__esModule") &&
+      ! (__esSymbol && hasOwn.call(exports, __esSymbol))) {
```
jdalton: This checks for the property "__esModule" or the `esSymbol`.

lib/runtime.js
```diff
@@ -125,13 +112,39 @@ exports.enable = function (Module) {
...
+
+    if (__esSymbol) {
+      exports[__esSymbol] = true;
+    } else {
```
jdalton: Prefer the symbol use for a 30-40ms perf win :tada:

lib/entry.js
```diff
@@ -94,100 +80,42 @@ Ep.runModuleGetters = function (module) {
...
-
   // Lazily-initialized object mapping parent module identifiers to parent
   // module objects whose setters we might need to run.
   var relevantParents;
 
   // Take snapshots of setter.parent.exports for any setters that we are
   // planning to call, so that we can later determine if calling the
   // setters modified any of those exports objects.
-  forEachSetter(function (setter, value, name) {
+  forEachSetter(module, this, function (setter, name, value) {
```
benjamn: Why not keep the `value, name` parameter ordering throughout this code?

lib/entry.js
```diff
@@ -94,100 +80,42 @@ Ep.runModuleGetters = function (module) {
...
     relevantParents = relevantParents || Object.create(null);
     relevantParents[setter.parent.id] = setter.parent;
+
+    // The param order for setters is `value` then `name` because the `name`
+    // param is only used by namespace exports.
     setter.call(module, setter.last[name] = value, name);
```
benjamn: Hmm, it looks like we might actually end up passing `NAN` or `UNDEFINED` to the setter function, instead of the real value (`NaN` or `void 0`);

lib/entry.js
jdalton: common setter order is `name` of the thing you want to set then the `value` of the thing to set. The `setter` bit is a backwards from convention so it gets a comment just below.

lib/entry.js
```diff
@@ -94,100 +80,42 @@ Ep.runModuleGetters = function (module) {
...
+  if (getter) {
     try {
       // Update module.exports[name] with the current value so that CommonJS
       // require calls remain consistent with module.importSync.
-      return module.exports[name] =
-        this.getters[name].call(module);
-
-    } catch (e) {
-      // If the getter threw an exception, avoid updating module.exports
-      // and return undefined.
```
benjamn: Let's keep this comment.

lib/entry.js
jdalton: Yes, this shouldn't effect the getter results so should be fine.

jdalton: The user shouldn't ever see the `NAN` or `UNDEFINED` objects.

lib/entry.js
```diff
@@ -94,100 +80,42 @@ Ep.runModuleGetters = function (module) {
...
-    }
+      module.exports[name] = getter();
```
benjamn: I feel like `runGetter` should return the gotten value, even if we're not using it currently. I'm aware this isn't completely rational, but it seems like something that might be surprising later.

lib/entry.js
jdalton: Convention like `element.setAttribute(name, value)` or `Object.defieProperty(object, name, descriptor)`, etc.

lib/entry.js
```diff
@@ -209,4 +136,74 @@ Ep.runModuleSetters = function (module) {
...
+
+  if (value !== value) {
+    value = NAN;
+  } else if (typeof value === "undefined") {
+    value = UNDEFINED;
+  }
+
+  setter.last = setter.last || Object.create(null);
+
+  if (setter.last[name] !== value) {
```
benjamn: I think we should update `setter.last[name] = value` here instead of in the callback, unless there's a good reason to split that apart.

lib/entry.js
```diff
@@ -209,4 +136,74 @@ Ep.runModuleSetters = function (module) {
...
+    // Only invoke the callback if we have not called this setter
+    // (with a value of this name) before, or the current value is
+    // different from the last value we passed to this setter.
+    return callback(setter, name, value);
```
benjamn: I think the third argument should be the original `value` passed into this function, even if it was replaced by `NAN` or `UNDEFINED` for the `setter.last[name]` logic.

lib/entry.js
jdalton: You can always add it back when it's used. Adding dead code for the feels isn't really a good thing.

lib/entry.js
jdalton: Ok cool., it was done in the [callback before](https://github.com/benjamn/reify/blob/b5f78d92968562a8497beebb61e8e0417693d10a/lib/entry.js#L184-L188) (so I just carried it over).

lib/entry.js
jdalton: Ok, fix it up.

lib/entry.js
jdalton: Ok. I thought it was stating the `modules.export` is only set in the `try` bit and nowhere else in the function.

PR #140, opened 2017-04-20
Make pseudo private methods private helpers.

jdalton:
Went ahead and made the pseudo private methods private helpers since we already have private helpers there's no need to bolt on stuff.

### DIFF
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -69,14 +69,7 @@ module.exports = class ImportExportVisitor extends Visitor {
         }
       }
 
-      bodyInfo.body = null;
-      bodyInfo.parent = null;
-      bodyInfo.hoistedExportsMap = null;
-      bodyInfo.hoistedExportsString = null;
-      bodyInfo.hoistedImportsString = null;
-      bodyInfo.insertCharIndex = null;
-      bodyInfo.insertNodeIndex = null;
-      bodyInfo.needToAddUseStrictDirective = null;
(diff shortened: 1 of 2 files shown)


lib/import-export-visitor.js
```diff
@@ -70,13 +70,14 @@ module.exports = class ImportExportVisitor extends Visitor {
...
 
       bodyInfo.body = null;
-      bodyInfo.parent = null;
       bodyInfo.hoistedExportsMap = null;
       bodyInfo.hoistedExportsString = null;
       bodyInfo.hoistedImportsString = null;
       bodyInfo.insertCharIndex = null;
       bodyInfo.insertNodeIndex = null;
       bodyInfo.needToAddUseStrictDirective = null;
+      bodyInfo.parent._bodyInfoByName = null;
```
jdalton: This is the last bit of pseudo private stuff, `parent._bodyInfoByName`. I noticed while we cleared body info on the visitor we didn't clear it on the ast node. So I went ahead and did that too.

benjamn: Cool. Might be worth actually deleting the property, since it's attached to a shared AST node, but nulling it is probably fine.

jdalton: Updated with `delete`.

jdalton: Things can actually be simplified since the bodyInfo is discarded when te bodyInfos array is cleared the nulling isn't needed and only the delete remains.

PR #143, opened 2017-04-20
Remove dead code from `getBlockBodyInfo`.

jdalton:
I think this is dead code but I wanted to check.

### DIFF
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -9,7 +9,6 @@ const Visitor = require("./visitor.js");
 const carriageReturnCode = "\r".charCodeAt(0);
 const exportDefaultPrefix = 'module.export("default",exports.default=(';
 const exportDefaultSuffix = "));";
-const hasOwn = Object.prototype.hasOwnProperty;
 
 module.exports = class ImportExportVisitor extends Visitor {
   finalizeHoisting() {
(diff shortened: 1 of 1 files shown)


lib/import-export-visitor.js
```diff
@@ -415,17 +414,11 @@ function getBlockBodyInfo(visitor, path) {
     // If parent is a Program, we may need to add a "use strict"
     // directive to enable const/let in Node 4.
     needToAddUseStrictDirective = visitor.enforceStrictMode;
-
     break;
 
   case "BlockStatement":
     body = parent.body;
-
```
jdalton: Because it's a block statement won't it always have a parent?
If not what test can we create to cause it to fail?

lib/import-export-visitor.js
```diff
@@ -415,17 +414,11 @@ function getBlockBodyInfo(visitor, path) {
...
-    if (hasOwn.call(parent, "start")) {
```
benjamn: The `BlockStatement`s we create in the `default` case below may not have a `.start` property, so this code was trying to be defensive about that.

jdalton: What's the default block used for?

jdalton: It looks like it's to replace blocks like `WhileStatement`, `IfStatement`, and `ForInStatement`.
Is there a reason to not have a start?

benjamn: It handles braceless statements like `if (condition) import "module"` by wrapping them with braces, since the block will need to contain both the variable declaration and the `module.importSync` call.

If we wanted to get really clever, I suppose we could include the `module.importSync` in the variable declaration, as the right-hand initializer for some dummy variable, but that might be more trouble than it's worth.

jdalton: Could we use the start position of the to-be-replaced node as the start of the default block?

jdalton: I see that with the source mod of adding `{` and `}` but how does the empty `block` when modifying AST help with that?

lib/import-export-visitor.js
```diff
@@ -402,48 +401,40 @@ function getBlockBodyInfo(visitor, path) {
...
 
   default:
-    bodyName = path.getName();
-
     const block = {
       type: "BlockStatement",
-      body: []
+      body: [],
+      start: node.start,
+      end: node.end + 2
```
jdalton: To avoid the `start` issue I added a `start` and `end` which aligns with what output would generate via [astexplorer](https://astexplorer.net/).

jdalton: Now though later when the block is expanded the start/end aren't updated so I'll look at that.

benjamn: I wouldn't worry about making sure all the `.start` and `.end` values are correct at the end. That information is pretty much destroyed by any AST modification.

jdalton: Should it be correct for when `modifyAST` is true though? I don't work much with AST mods is the expectation that things line up (start/ends/etc)?

benjamn: Since we're using the start/end information to modify the original source, they should point to a correct position in the source code, not the final generated code. If you think about it that way, the start/end values of the final AST aren't wrong after modifying the AST; they just still refer to the original source.

jdalton: Ah, ok! Then this PR is good to go!

Issue #141, opened 2017-04-20
Avoid double reads of files.


jdalton:
The current compile hook ties into `Module.prototype._compile` but by the time that method is called [the original source has been read](https://github.com/nodejs/node/blob/d8965d5b0ec57d628e11bba17d0ff14aad28bf78/lib/module.js#L591-L594) with `fs.readFileSync` it then checks the cache and does another `fs.readFileSync` if the cache exists. Instead of checking for cache at the point of `Module.prototype._compile` it could use `Module._extensions[".js"]` and friends to check for a cache upfront before reading the file the first time and avoid double reads which should significantly improve cache load times.

jdalton:
Moved to #144.

PR #144, opened 2017-04-24
Avoid double reads when loading cached files.

jdalton:
This PR addresses #141 by avoiding double reading of cache files, safely *(with fallbacks)* uses optimized internal helpers for reading and stating files, reduces memory a bit by storing less package.json data, defers creating cache directories until cache files are being written to, and reduces unnecessary crawling by a level.

### DIFF
--- a/node/utils.js
+++ b/node/utils.js
@@ -1,12 +1,37 @@
 "use strict";
 
 const createHash = require("crypto").createHash;
-const dynRequire = module.require ? module.require.bind(module) : __non_webpack_require__;
 const fs = require("fs");
 const path = require("path");
+const zlib = require("minizlib");
 
-const DEFAULT_CACHE_DIR = ".reify-cache";
+const DEFAULT_GZIP_CONFIG = {
+  level: 9
+};
+
+const DEFAULT_PKG_CONFIG = {
+  "cache-directory": ".reify-cache",
+  parser: undefined
+};
+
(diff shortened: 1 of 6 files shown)


benjamn:
Awesome!

jdalton:
It's more than a 50% win taking the load time down to that of normal cjs.

jdalton:
So that's mid 100 milliseconds for loading 642 modules.

node/utils.js
```diff
@@ -27,193 +49,251 @@ const reifyVersion = (() => {
...
-      ! checkReify(pkg, "peerDependencies") &&
       // Use case: a package.json file may have "reify" in its
       // "devDependencies" section because it expects another package or
       // application to enable reification in production, but needs its
       // own copy of the "reify" package during development. Disabling
       // reification in production when it was enabled in development
       // would be dangerous in this case.
-      ! checkReify(pkg, "devDependencies")) {
+      ! checkReify(pkg, "devDependencies") &&
+      ! checkReify(pkg, "peerDependencies")) {
```
benjamn: Nice catch (that the comment applies to both `dev` and `peer` dependencies).

jdalton: Yep. It was actually because `peerDependencies` are used less so faster fail by checking devDeps first.

jdalton:
Updated with caching as gzipped files too.

PR #146, opened 2017-04-24
Implement Export{Default,Namespace}Specifier parsing ourselves.

benjamn:
Related: https://github.com/ternjs/acorn/pull/541

### DIFF
--- a/lib/parsers/acorn-extensions.js
+++ b/lib/parsers/acorn-extensions.js
@@ -1,0 +1,191 @@
+"use strict";
+
+const tt = require("acorn").tokTypes;
+const hasOwn = Object.prototype.hasOwnProperty;
+
+exports.fix = function (parser) {
+  // It's not Reify's job to enforce strictness.
+  parser.strict = false;
+
+  // Tolerate recoverable parse errors.
+  parser.raiseRecoverable = noopRaiseRecoverable;
+
+  // Lookahead method.
+  parser.lookAhead = lookAhead;
+
+  // Export-related modifications.
+  parser.parseExport = parseExport;
+  parser.isExportDefaultSpecifier = isExportDefaultSpecifier;
+  parser.parseExportSpecifiersMaybe = parseExportSpecifiersMaybe;
+  parser.parseExportFromWithCheck = parseExportFromWithCheck;
+  parser.parseExportFrom = parseExportFrom;
+  parser.checkExport = checkExport;
+  parser.shouldParseExportDeclaration = shouldParseExportDeclaration;
+};
+
+function noopRaiseRecoverable() {}
+
+function parseExport(node, exports) {
+  this.next();
+  if (this.type === tt.star) {
+    const specifier = this.startNode();
+    this.next();
+    if (this.eatContextual("as")) {
+      // export * as ns from '...'
+      specifier.exported = this.parseIdent(true);
+      node.specifiers = [
+        this.finishNode(specifier, "ExportNamespaceSpecifier")
+      ];
+      this.parseExportSpecifiersMaybe(node);
+      this.parseExportFromWithCheck(node, exports, true);
+    } else {
+      // export * from '...'
+      this.parseExportFromWithCheck(node, exports, true);
+      return this.finishNode(node, "ExportAllDeclaration");
+    }
+  } else if (this.isExportDefaultSpecifier()) {
+    // export def from '...'
+    const specifier = this.startNode();
+    specifier.exported = this.parseIdent(true);
+    node.specifiers = [
+      this.finishNode(specifier, "ExportDefaultSpecifier")
+    ];
+    if (this.type === tt.comma &&
+        this.lookAhead(1).type === tt.star) {
+      // export def, * as ns from '...'
+      this.expect(tt.comma);
+      const specifier = this.startNode();
+      this.expect(tt.star);
+      this.expectContextual("as");
+      specifier.exported = this.parseIdent(true);
+      node.specifiers.push(
+        this.finishNode(specifier, "ExportNamespaceSpecifier")
+      );
+    } else {
+      // export def, { x, y as z } from '...'
+      this.parseExportSpecifiersMaybe(node);
+    }
+    this.parseExportFromWithCheck(node, exports, true);
+  } else if (this.eat(tt._default)) {
+    // export default ...
+    this.checkExport(exports, "default", this.lastTokStart);
+    let isAsync;
+    if (this.type === tt._function || (isAsync = this.isAsyncFunction())) {
+      let fNode = this.startNode();
+      this.next();
+      if (isAsync) this.next();
+      node.declaration = this.parseFunction(fNode, "nullableID", false, isAsync);
+    } else if (this.type === tt._class) {
+      let cNode = this.startNode();
+      node.declaration = this.parseClass(cNode, "nullableID");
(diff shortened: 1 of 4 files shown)


lib/parsers/acorn.js
```diff
@@ -1,6 +1,7 @@
 "use strict";
 
 const acorn = require("acorn");
+const fixParser = require("./acorn-extensions.js").fix;
```
jdalton: Maybe a more descriptive name than `fix`. It looks like the `acorn-extensions` could be a place to host more than 1 extension.

lib/parsers/acorn-extensions.js
```diff
@@ -0,0 +1,191 @@
...
+  }
+  return this.finishNode(node, "ExportNamedDeclaration");
+}
+
+function lookAhead(n) {
+  const old = Object.assign(Object.create(null), this);
+  while (n-- > 0) this.nextToken();
+  const copy = Object.assign(Object.create(null), this);
+  Object.assign(this, old);
+  return copy;
```
jdalton: What's the copying about?

lib/parsers/acorn-extensions.js
```diff
@@ -0,0 +1,191 @@
...
+  if (this.eatContextual("from")) {
+    node.source = this.type === tt.string
+      ? this.parseExprAtom()
+      : this.unexpected();
+  } else {
+    if (node.specifiers) {
+      // check for keywords used as local names
+      for (let i = 0; i < node.specifiers.length; i++) {
+        const local = node.specifiers[i].local;
+        if (local && (this.keywords.test(local.name) ||
```
jdalton: what's the possible values of `local`? If it's either an object or `undefined` we should be explicit about the object check with `typeof local === "object" && local !== null` as v8 (and I guess others) favor that over a bool check.

lib/parsers/acorn-extensions.js
```diff
@@ -0,0 +1,191 @@
...
+                      this.reservedWords.test(local.name))) {
+          this.unexpected(local.start);
+        }
+      }
+    }
+
+    if (expect) {
+      this.unexpected();
+    } else {
```
jdalton: expect the unexpected 🤣

lib/parsers/acorn-extensions.js
```diff
@@ -0,0 +1,191 @@
...
+      node.source = null;
+    }
+  }
+
+  this.semicolon();
+}
+
+function checkExport(exports, name, pos) {
+  if (!exports) return;
```
jdalton: Similar object check for `exports`.

lib/parsers/acorn-extensions.js
```diff
@@ -0,0 +1,191 @@
...
+        this.lookAhead(1).type === tt.star) {
+      // export def, * as ns from '...'
+      this.expect(tt.comma);
+      const specifier = this.startNode();
+      this.expect(tt.star);
+      this.expectContextual("as");
+      specifier.exported = this.parseIdent(true);
+      node.specifiers.push(
+        this.finishNode(specifier, "ExportNamespaceSpecifier")
+      );
```
jdalton: Just curious does the execution of the `this.expect` stuff just above matter? I noticed one was before the `const specifier =....`.

jdalton:
This is mostly all magic to me 😸

lib/parsers/acorn-extensions.js
benjamn: We're taking a snapshot of the parser so that we can roll it back after advancing its state to look at later tokens.

lib/parsers/acorn-extensions.js
benjamn: Yes, `this.expect(...)` modifies `this.start` (by calling `this.eat(...)` which calls `this.next()`), which [is used by `this.startNode`](https://github.com/ternjs/acorn/blob/1ecf4f737df99592372e517a8f7d09a6c125bc9f/src/node.js#L23) to set the `.start` location of the new node, so the relative ordering of `this.expect(...)` and `this.startNode()` is important.

lib/parsers/acorn-extensions.js
benjamn: I'll do you one better: this whole loop can be eliminated (soon)!

benjamn:
Thanks for the feedback! I think you're gonna like the follow-up commits I just pushed.

lib/parsers/acorn-extensions.js
```diff
@@ -0,0 +1,169 @@
...
+  while (n-- > 0) this.nextToken();
+  try {
+    return callback.call(this);
+  } finally {
+    Object.assign(this, old);
+  }
+}
+
+function peekNextType(parser) {
+  return parser.withLookAhead(1, () => parser.type);
```
benjamn: I really like this new interface, personally.

lib/parsers/acorn-extensions.js
jdalton: 😵

lib/parsers/acorn-extensions.js
jdalton: It's a shallow snapshot so not too bad.

lib/parsers/acorn.js
```diff
@@ -12,8 +12,18 @@ exports.options = {
...
+  if (acornExtensions === null) {
+    acornExtensions = require("./acorn-extensions.js");
+  }
+
   const parser = new acorn.Parser(exports.options, code);
-  fixParser(parser);
+
+  acornExtensions.enableAll(parser);
+
   return parser.parse();
```
jdalton: What's the lazy loading of `acorn` and `acorn-extensions` do?

lib/parsers/acorn-extensions.js
```diff
@@ -106,23 +106,33 @@ function parseExport(node, exports) {
...
+// back to its original state and returns whatever the callback returned.
+function withLookAhead(n, callback) {
   const old = Object.assign(Object.create(null), this);
   while (n-- > 0) this.nextToken();
-  const copy = Object.assign(Object.create(null), this);
-  Object.assign(this, old);
-  return copy;
+  try {
+    return callback.call(this);
+  } finally {
```
jdalton: What's a typical callback look like?

jdalton: If it's just `() => parser.type` you can optimize the callback call without the `this` binding and just do
```
return callback();
```

lib/parsers/acorn.js
benjamn: Hmm, probably not much if we always use the parser when we import this module (which I think we do). Are you concerned it might hurt?

lib/parsers/acorn-extensions.js
jdalton: Ah I see `isCommaOrFrom` uses `this`. No worries.

benjamn: There's another case (`isCommaOrFrom`) where we use `this`.

Any preference between these two options?
```js
return callback.call(this);
// or
return callback(this);
```
?

jdalton: Yes! In event loops it's always sad to see folks default to applying a `this` binding because many times it's not used and does add a cost to invocation. So if you could pass in the `this` as the first arg that is super nice. It's what I would recommend for event systems too!

jdalton:
👍

Issue #139, opened 2017-04-19
Use a unique reference for `module` in case the user creates their own local identifier.


jdalton:
We should create a unique reference for the `module` object in case the user creates their own `module` identifier.

jdalton:
Moved to #164.

PR #147, opened 2017-04-25
Use a smarter importExportRegExp to avoid recompilation.

benjamn:
The compiler uses `importExportRegExp` to check quickly whether a string of code contains anything that looks like an `import` or `export` keyword, so it can avoid parsing, AST transformation, etc. when there's clearly no work to do.

A particularly important source of false positives was that the compiler itself generates `module.importSync` and `module.export` calls to replace `import` and `export` declarations. Since `/\b(?:im|ex)port\b/.test("module.export(...)")` is true, the compiler would go ahead and recompile code that contained `module.export` expressions (`module.import` is no longer as much of a problem since we renamed it to `module.importSync` in #85).

By changing `/\b(?:im|ex)port\b/` to `/(?:^|[^.])\b(?:im|ex)port\b/`, we can exclude matches that are preceded by a `.` character, which means the compiler should never recompile code that it previously compiled (unless there were other false positives that were not replaced, which is beyond the scope of the quick-and-dirty `importExportRegExp` check).

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -12,7 +12,13 @@ const AV = require("./assignment-visitor.js");
 const assignmentVisitor = new AV;
 
 const shebangRegExp = /^#!.*/;
-const importExportRegExp = /\b(?:im|ex)port\b/;
+
+// Matches any import or export identifier as long as it's not preceded by
+// a `.` character (to avoid matching module.export, for example). Because
+// Reify replaces all import and export declarations with module.import
+// and module.export calls, this logic should prevent the compiler from
+// ever having to recompile code it has already compiled.
+const importExportRegExp = /(?:^|[^.])\b(?:im|ex)port\b/;
 
 exports.compile = function (code, options) {
   code = code.replace(shebangRegExp, "");


lib/compiler.js
```diff
@@ -12,7 +12,13 @@ const AV = require("./assignment-visitor.js");
...
 
 const shebangRegExp = /^#!.*/;
-const importExportRegExp = /\b(?:im|ex)port\b/;
+
+// Matches any import or export identifier as long as it's not preceded by
+// a `.` character (to avoid matching module.export, for example). Because
+// Reify replaces all import and export declarations with module.import
+// and module.export calls, this logic should prevent the compiler from
+// ever having to recompile code it has already compiled.
+const importExportRegExp = /(?:^|[^.])\b(?:im|ex)port\b/;
```
jdalton: I like where this is going. Could this get a bit more specific. What can come before the `import` or `export` keyword? Is it just whitespace?

benjamn: I'm guessing it's whitespace, a `;`, or the beginning of the string (that is, nothing).

benjamn: Oh also a `)`, as in `if(condition)import "foo"`.

jdalton: And `{`? But that's also depending on if nested imports is on. Smth to keep in mind when it's configurable is that the regexp can toggle depending.

jdalton: There's also a balance between a quick check and the cost of a complex regexp so I'd side on simplicity over complex.... there are regexp helpers that help point out the complexity of a regexp.

benjamn: I would imagine the current version is pretty simple: checking that the previous character is not a `.` should be faster than checking that it is one of several different options. The implementation might not take full advantage of that, but it seems straightforward to get it right.

jdalton: https://regex101.com/ is a nice one.

benjamn: And the benefit of avoiding reparsing and recompiling is pretty huge!

jdalton: This makes the steps ~x3 more for a failed match or match of just exports. Maybe a combo of a light and then a heavier regexp... looking into it.

jdalton: It looks like its better to do two regexes.
The original light one of  `/\b(?:im|ex)port\b/` then one for `module\.(im|ex)port`

jdalton: So the light one will filter out most cjs and then the secondary one will filter out transpiled sources. Though in the common case of reify it wont be processing sources that it's transpiled since they will be read from the cache.

jdalton: `module\.(?:import(?:Sync)?|export)`

benjamn: Here's a microbenchmark for positive vs. negative: https://jsperf.com/negative-vs-positive-regexp-lookbehind (tl;dr it doesn't make much difference, and `[^.]` is more conservative/safer, I think).

benjamn: Also remember (per #139) that we can't necessarily rely on the `module` identifier.

benjamn: I don't see how the two `RegExp` approach makes sense, because you'd have to save all the matches from the first regular expression in order to check them against the second one.

jdalton:
Ah nice, ok looks good!

PR #149, opened 2017-04-27
Cleanup acorn-extensions.

jdalton:
This PR cleans up the acorn-extensions.

* Removed unneeded lazy loading logic of `acorn` and `acorn-extensions`
* Removes duplicated functionality that's already provided by acorn
* Overwrites `parser.checkExports` to remove checks which removes checks from `parseExportSpecifiers` (which means our lack of checks is consistent 😃 )
* It removes a rogue trailing semicolon from exported functions/classes (I should patch acorn/babylon)
* It avoids extending the parser with non-standard methods (keeps them helpers instead)
* It splits out the massive `parseExport` into more manageable chunks.

### DIFF
--- a/lib/parsers/acorn-extensions.js
+++ b/lib/parsers/acorn-extensions.js
@@ -3,8 +3,13 @@
 const tt = require("acorn").tokTypes;
 
 exports.enableAll = function (parser) {
-  exports.enableTolerance(parser);
   exports.enableExportExtensions(parser);
+  exports.enableTolerance(parser);
+};
+
+exports.enableExportExtensions = function (parser) {
+  parser.checkExports = checkExports;
+  parser.parseExport = parseExport;
 };
 
 exports.enableTolerance = function (parser) {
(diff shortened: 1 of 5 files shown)


test/output-tests.js
```diff
@@ -28,8 +28,8 @@ Object.keys(files).forEach((absPath) => {
 describe("output", () => {
   function check(data) {
     const code = compile(data.actual).code;
-    // Trim blank lines and trailing whitespace.
-    const actual = code.replace(/^ +$/gm, "").trimRight();
+    // Consolidate semicolons then trim blank lines and trailing whitespace.
+    const actual = code.replace(/;{2,}/g, ";").replace(/^ +$/gm, "").trimRight();
```
jdalton: While I removed the rogue semicolon in acorn (due to the acorn-extensions) it still exists in babylon.

test/output/anon-class/expected.js
```diff
@@ -2,4 +2,4 @@
   constructor(value) {
     this.value = value;
   }
-}));;
+}));
```
jdalton: ^that rogue semicolon

lib/parsers/acorn-extensions.js
```diff
@@ -15,154 +15,180 @@ exports.enableTolerance = function (parser) {
...
+    node.declaration = parser.parseClass(classNode, "nullableID");
   } else {
-    // export { x, y as z } [from '...']
-    node.declaration = null;
-    node.specifiers = this.parseExportSpecifiers(exports);
-    this.parseExportFrom(node, exports);
+    node.declaration = parser.parseMaybeAssign();
   }
-  return this.finishNode(node, "ExportNamedDeclaration");
+  parser.semicolon();
```
jdalton: That ^ removes the rogue semicolon.

lib/parsers/acorn-extensions.js
```diff
@@ -15,154 +15,180 @@ exports.enableTolerance = function (parser) {
...
-    }
-  }
-
-  this.semicolon();
+function peekNextType(parser) {
+  return withLookAhead(parser, 1, () => parser.type);
 }
 
-function shouldParseExportDeclaration() {
-  return this.type.keyword === "var" ||
```
jdalton: `shouldParseExportDeclaration` is acorn's `this.shouldParseExportStatement`.

test/output/anon-class/expected.js
benjamn: Yeah, I've noticed that little sucker!

lib/parsers/acorn-extensions.js
benjamn: The `shouldParseExportDeclaration` is more accurate than `shouldParseExportStatement`, though.

lib/parsers/acorn-extensions.js
```diff
@@ -15,154 +15,180 @@ exports.enableTolerance = function (parser) {
...
+
+  parseExportFrom(parser, node);
+  return parser.finishNode(node, "ExportNamedDeclaration");
 }
 
-function isCommaOrFrom(parser) {
-  return parser.type === tt.comma ||
-    (parser.type === tt.name &&
-     parser.value === "from");
+function parseExportNamespaceAs(parser, node, specifier, exports) {
```
benjamn: Let's call this `parseExportNamespaceSpecifier`.

lib/parsers/acorn-extensions.js
```diff
@@ -15,154 +15,180 @@ exports.enableTolerance = function (parser) {
...
+  } else {
+    parser.semicolon();
+  }
+  return parser.finishNode(node, "ExportNamedDeclaration");
 }
 
-function isExportDefaultSpecifier() {
-  return this.type === tt.name &&
-    this.withLookAhead(1, isCommaOrFrom);
+function parseExportNamespace(parser, node, exports) {
```
benjamn: Let's call this `parseExportNamespaceSpecifiersAndSource`, since there can be other specifiers, and we're also parsing the `from "..."` part here.

lib/parsers/acorn-extensions.js
```diff
@@ -15,154 +15,180 @@ exports.enableTolerance = function (parser) {
...
+  // export def, * as ns [, { x, y as z }] from '...';
+  parseExportSpecifiersMaybe(parser, node);
+
+  parseExportFrom(parser, node);
+  return parser.finishNode(node, "ExportNamedDeclaration");
 }
 
-function peekNextType(parser) {
-  return parser.withLookAhead(1, () => parser.type);
+function parseExportImportSpecifiers(parser, node, exports) {
```
benjamn: Let's call this `parseExportSpecifiersAndSource`.

lib/parsers/acorn-extensions.js
```diff
@@ -15,154 +15,180 @@ exports.enableTolerance = function (parser) {
...
-    return callback(this);
-  } finally {
-    Object.assign(this, old);
+function parseExportFrom(parser, node) {
+  parser.expectContextual("from");
+  node.source = parser.type === tt.string ? parser.parseExprAtom() : null;
+  parser.semicolon();
+}
+
+function parseExportImportDefault(parser, node, exports) {
```
benjamn: How about `parseExportDefaultSpecifiersAndSource`, like `parseExportNamespaceSpecifiersAndSource`?

lib/parsers/acorn-extensions.js
```diff
@@ -15,154 +15,180 @@ exports.enableTolerance = function (parser) {
...
+
+  if (node.declaration.type === "VariableDeclaration") {
+    parser.checkVariableExport(exports, node.declaration.declarations);
+  } else {
+    exports[node.declaration.id.name] = true;
+  }
+  return parser.finishNode(node, "ExportNamedDeclaration");
+}
+
+function parseExportDefault(parser, node, exports) {
```
benjamn: The convention in Acorn is to include the full AST node type in the function name, if the function is responsible for parsing that node type. So I would call this `parseExportDefaultDeclaration`.

lib/parsers/acorn-extensions.js
```diff
@@ -15,154 +15,180 @@ exports.enableTolerance = function (parser) {
...
+function isCommaOrFrom(parser) {
+  return parser.type === tt.comma || parser.isContextual("from");
+}
+
+function isExportDefaultSpecifier(parser) {
+  return parser.type === tt.name &&
+    withLookAhead(parser, 1, isCommaOrFrom);
+}
+
+function parseExportDeclaration(parser, node, exports) {
```
benjamn: Let's call this `parseExportNamedDeclaration`, since that's what it does.

lib/parsers/acorn-extensions.js
jdalton: What's the significance of the `AndSource` bit?

lib/parsers/acorn-extensions.js
jdalton: It's really only processing the "As" case though. Is there a more specific name than the generic parseExportNamespaceSpecifier?

lib/parsers/acorn-extensions.js
jdalton: Is the `AndSource` an acorn convention too?

benjamn: No, this is a convention that we're introducing, since Acorn doesn't extract any functions for the `node.source = ...` code.

Come to think of it, since we added `parseExportFrom`, I think we should rename that to `parseExportSource` for consistency, since "from" makes a weird noun.

lib/parsers/acorn-extensions.js
jdalton: Nm, `parseExportNamespaceSpecifiers` is technically correct.

jdalton:
Updated.

PR #151, opened 2017-04-27
Rename `force` option to `sourceType`.

jdalton:
This renames the `force` option to `sourceType` which aligns [with babel/babylon](https://github.com/babel/babylon/pull/449).


jdalton:
I'm going merge since this isn't super controversial and aligns with config that's recognizable from other projects like Babel, TypeScript, and ESLint.

PR #152, opened 2017-05-01
Adds support for multiple versions of reify to be used by different packages.

jdalton:
This PR adds support for multiple versions of reify to be used by different packages.

This applies wrappers based on the package's specified reify version range. The version range for wrappers is pulled from the packages dependencies, peerDependencies, or devDependencies and a wrapper wrapper is chosen which maximally satisfies the range.

This also avoids extending the `Module.prototype`.
All runtime extensions are done on `module` instances.


jdalton:
This applies wrappers following semver that means for pre 1.0 wrappers will be stored as 0.x and for post 1.0 they will be stored with just their major release. For example if a wrapper for v0.8.4 is applied but later there is a v0.8.5 wanting to be applied then the v0.8.5 will replace the v0.8.4. However if a v0.9.0 is wanting to be applied, and no other 0.9 entry exists, it will create a new slot for that wrapper and use that for v0.9.

The version range for wrappers is pulled from the packages dependencies, peerDependencies, or devDependencies and a wrapper wrapper is chosen which maximally satisfies the range.

node/compile-hook.js
```diff
@@ -3,78 +3,111 @@
...
+}
+
+function getWrapper(func, filename) {
+  const pkgInfo = typeof filename === "string"
+    ? utils.getPkgInfo(path.dirname(filename))
+    : null;
+
+  if (pkgInfo !== null) {
+    const wrappers = func.reified.wrappers;
+    const versions = Object.keys(wrappers);
```
jdalton: If `Object.keys` is a perf concern we can maintain a separate array of versions. I haven't done a perf check of this stuff yet. But it's nothing that can't be addressed as needed.

jdalton:
I need to add support for `Module.prototype.load` and `Module.prototype.resolve` wrappers too.

jdalton:
This also avoids extending the `Module.prototype` for everything but the `Module.prototype._compile` hook. All the other runtime extensions are done on `module` instances.

lib/runtime.js
```diff
@@ -127,7 +64,70 @@ exports.enable = function (Module) {
...
+    // becomes
+    //
+    //   module.export("a", () => a);
+    //   var a = 1;
+    //   console.log(module.runModuleSetters(a += 3));
+    //
+    // This ensures module.runModuleSetters runs immediately after the
+    // assignment, and does not interfere with the larger computation.
+    return valueToPassThrough;
+  };
```
benjamn: Since these instance methods only refer to `this` and not `module`, I think we can define them at the top level (outside of the `enable` function) and just copy them onto the `module` instance.

node/compile-hook.js
```diff
@@ -80,6 +83,7 @@ createWrapperManager(exts, ".mjs");
 addWrapper(Mp._compile, function(func, content, filename) {
   const options = Object.assign({ filename }, compileOptions);
   options.cacheKey = () => getCacheKey(filename);
+  runtime.enable(this);
```
benjamn: I wonder if we could make Reify work in any CommonJS environment by calling `require("reify/lib/runtime").enable(module)` at the very top of compiled modules, rather than here?

lib/runtime.js
jdalton: Would you apply them as:
```js
Object.assign(module, require('./runtime.js'))
```
?

node/compile-hook.js
jdalton: This is just for adding the runtime helpers. This doesn't opt-in arbitrary code into the reify compiler.

benjamn:
I'm uneasy about the interaction of multiple copies of Reify, especially because it looks like they'll each have their own `Entry` implementation and `entryMap` object. Will a module using one version of Reify be able to import, or be imported by, a module using a different version? Until now, the shared `entryMap` has been implicitly assumed, and I can't say I fully understand the consequences of having more than one. Have you thought this through?

jdalton:
Multiple versions of reify should see each other as cjs. They're runtimes are independent of each other.

benjamn:
But the parent module needs to add setters to the `Entry` object of the imported child module, right? Doesn't that mean the child can't just be treated as CJS?

jdalton:
Ideally the different versions aren't aware of each other. They should be isolated.
If you're concerned a unit test are mock setup of a scenario would help.

benjamn:
Sure, we should work up a test of this stuff. My previous comment seems like it will be a blocker, though. The `module.importSync` API is basically subscribing to the exports of the child module, so the child has to be able to broadcast those changes to its parents. That may mean stitching together different `Entry` implementations, but it's something I think we'll have to address.

jdalton:
I'll create a playground repo so you can kick stuff around.

benjamn:
To be clear, I totally agree we have to figure out versioning so that we don't just let the first copy of Reify "win." I'm just looking for the cracks that might open up as we go down that path.

jdalton:
I've created a mock scenario at [reify-version-check](https://github.com/jdalton/reify-version-check). Once you clone the repo you can just do `node index.js`. Feel free to kick things around.

jdalton:
I've updated the implementation to upgrade and migrate entry maps when wrappers are upgraded.
That said I've noticed this common pattern:

A package, "package B", will use Reify. To do this its package entry, `index.js`, is a small CJS file that bootstraps reify and loads the main ESM module.
```js
require("reify");
module.exports = require("./main.js");
```
The CJS bootstrap nixes live bindings. So if it was consumed by another ESM package:
```js
import { x } from "b"
```
The `x` specifier would not be live because the entry to "b" is the CJS `index.js`.

In cases where packages share the same range reify range they'll also share the same entry cache and if you were to load an esm module from within another esm module then live bindings would be preserved. However I don't see that as being the common pattern (the cjs `index.js` is).

That said. Live bindings will work within a package just fine as long as its esm loading esm so that's still nice.

In cases where a package opts in to a version of reify and tries to load raw ESM of another package that opts-in to a different range then it's treated as trying to load esm without a reify transpiler (so it just errors out which makes sense).

jdalton:
Rebased.

lib/runtime.js
```diff
@@ -3,17 +3,16 @@
...
 
-exports.enable = function (Module) {
-  var Mp = Module.prototype;
-
-  if (typeof Mp.importSync !== "function" ||
-      typeof Mp.export !== "function") {
-    Mp.runModuleSetters = runModuleSetters;
-    Mp.importSync = moduleImport;
-    Mp.export = moduleExport;
+exports.enable = function (mod) {
```
benjamn: How would you feel about allowing a constructor function, too?
```js
exports.enable = function (mod) {
  if (typeof mod === "function") mod = mod.prototype;
  ...
};
```
With that option, a caller who knows what they're doing could set `Module.prototype.{export,importSync,runModuleSetters}` for the entire process, overriding individual Reify versions. It would also be backwards compatible.

package.json
```diff
@@ -32,7 +32,8 @@
   },
   "dependencies": {
     "acorn": "~5.0.0",
-    "minizlib": "^1.0.3"
+    "minizlib": "^1.0.3",
```
benjamn: Why not just use the built-in `zlib`?

lib/entry.js
```diff
@@ -14,14 +14,16 @@ function Entry(exports) {
...
   // Getters for local variables exported from this module.
-  this.getters = new FastObject;
+  this.getters = Object.create(null);
+  // Setters for assigning to local variables in parent modules.
+  this.setters = Object.create(null);
 }
 
 var Ep = Entry.prototype;
 
+Ep.__proto__ = null;
```
benjamn: How does this help? Is it safe to set `__proto__` in all JS environments?

lib/runtime.js
jdalton: I'd want to avoid that touching global bits shouldn't be encouraged.

package.json
jdalton: `minilib` is 3x faster and not heavy to boot.

lib/entry.js
jdalton: `__proto__` is supported in IE 11+ and all other browsers. It won't hurt older ones though.
This avoids using ES6 `Object.setPrototypeOf`.

lib/runtime.js
jdalton: Since this is just bolting things on they could always pass `Module.prototype`. Our runtime stuff isn't really exposed though so I wouldn't want to support it for public mucking.

lib/entry.js
jdalton: Inheriting from `null` helps with our prop perf and aligns with our practice of using `Object.create(null)` style objects.

lib/runtime.js
benjamn: Oh that's true, passing `Module.prototype` would work just fine.

lib/entry.js
benjamn: Yeah, I would be even more worried about `Object.setPrototypeOf` without a polyfill. One more idea you've probably already considered… what about this?
```js
var Ep = Entry.prototype = Object.create(null);
Ep.constructor = Entry;
```

jdalton: `Object.setPrototypeOf` is OK in the Node 4+ code paths just as `Object.assign`, classes, `let`, and `const` are OK. Setting the `__proto__` is more aligned with the `Object.setPrototypeOf` use since it keeps the `prototype` object intact and just mods the inheritance of the `prototype` (`Object.setPrototypeOf` is the thing that `__proto__` became). There is no compat concern here.

jdalton:
Ok, let's flip the switch. There's lots of good things here and we can revise tweak as needed after the fact.

PR #154, opened 2017-05-02
Implement Module.isESModuleExports.

benjamn:
It was tempting to implement `Module.prototype.isESModule`, but it's also important to be able to determine if the `exports` object returned by a child module has a truthy `"__esModule"` or `__esSymbol` property, even if we don't have direct access to the child module's `module` object.

### DIFF
--- a/lib/runtime.js
+++ b/lib/runtime.js
@@ -56,7 +56,7 @@ exports.enable = function (Module) {
   // inside loops. The compiler generates these keys automatically (and
   // deterministically) when compiling nested import declarations.
   Mp.importSync = function (id, setters, key) {
-    setESModule(this);
+    setESModule(this.exports);
 
     var absoluteId = this.resolve(id);
 
@@ -88,7 +88,7 @@ exports.enable = function (Module) {
   // export statement. The keys of the getters object are exported names,
   // and the values are functions that return local values.
   Mp.export = function (getters) {
-    setESModule(this);
+    setESModule(this.exports);
 
     if (typeof getters === "object" && getters !== null) {
       Entry.getOrCreate(this.id).addGetters(getters);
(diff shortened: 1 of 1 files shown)


benjamn:
Should this be "private," i.e. `Module._isESModuleExports`?

jdalton:
I'm not sure what problem this is trying to solve.
Is this a way around creating a dynamic var reference for `module`?

benjamn:
It's so that external code can use the same `__esModule` logic that Reify uses, without having to know the details of `__esModule` vs. `__esSymbol`. I found myself checking for `__esModule` and `Symbol.for("__esModule")` in another library, then realized how fragile that was, so I thought Reify should export a helper to abstract away the details.

jdalton:
Can you give an example. I'm still a bit unclear.

benjamn:
Specifically, I'm trying to get Meteor to use Reify more extensively, by following the `jsnext:main` or `module` properties of `package.json` files. A beta user ran into [this bug](https://github.com/meteor/meteor/pull/8327#issuecomment-298723203) when a CJS module tried to `require` a package that had a `jsnext:main` entry point. If we had something like `Module.isESModuleExports`, then Meteor's implementation of `require` could potentially detect that case and return `exports.default` instead of just `exports`.

benjamn:
It's tempting just to revert the `jsnext:main`/`module` logic (recently added), but then we'd miss out on a worthwhile test of Reify in the wild.

jdalton:
Is this still needed?

Don't packages that use `jsnext:main/module` also ship cjs fallbacks?
Could they just use the cjs fallback?

benjamn:
I ended up mostly [reverting](https://github.com/meteor/meteor/pull/8327/commits/269e10caa6c72312e7032f4c23b4f06f974320d4) the changes I was trying to salvage with this API, so I think we can close this now. As a side note, this functionality is now sorta kinda exposed via `require("reify/lib/utils").getESModule(module.exports)`.

PR #158, opened 2017-05-07
Make require.extensions wrapper logic more defensive

benjamn:
Other libraries sometimes modify `require.extensions` to do their own compilation, so I think we should be as tolerant as we can of that.

In [`meteor-babel/register`](https://github.com/meteor/babel/blob/eb32b0907369bfab64ab649b7af1dbb719746dff/register.js#L36-L50) I noticed I had to copy the `.reified` property when I wrapped Reify's `require.extensions[".js"]` function: https://github.com/meteor/babel/commit/eb32b0907369bfab64ab649b7af1dbb719746dff

That seems like too much to ask of libraries that don't know about Reify. The most we can expect is for the other library to call the replaced function, like so:
```js
var defaultHandler = require.extensions[".js"];
require.extensions[".js"] = function(module, filename) {
  if (shouldNotTransform(filename)) {
    defaultHandler(module, filename);
  } else {
    module._compile(
      getBabelResult(filename).code,
      filename
    );
  }
};
```
I believe the most recent commit in this PR fixes this problem by using `wrapperFunc.reified` instead of `object[key].reified`, since `object[key]` might have changed, but I wanted to check with you (@jdalton) that this change seems safe.

### DIFF
--- a/node/wrappers.js
+++ b/node/wrappers.js
@@ -7,34 +7,60 @@ const isObject = require("../lib/utils.js").isObject;
 const utils = require("./utils.js");
 const reifySemVer = utils.getReifySemVer();
 const reifyVersion = reifySemVer.version;
+const hasOwn = Object.prototype.hasOwnProperty;
 
-function addWrapper(func, wrapper) {
-  const reified = func.reified;
-  if (typeof reified.wrappers[reifyVersion] !== "function") {
-    reified.versions.push(reifyVersion);
-    reified.wrappers[reifyVersion] = wrapper;
+function getSymbolForKey(key) {
+  return Symbol.for("reifyWrapperMap$" + key);
+}
+
+function addWrapper(object, key, wrapper) {
+  const map = ensureWrapperMap(object, key);
+  if (typeof map.wrappers[reifyVersion] !== "function") {
+    map.versions.push(reifyVersion);
+    map.wrappers[reifyVersion] = wrapper;
   }
 }
 
 exports.addWrapper = addWrapper;
 
-function createWrapperManager(object, key) {
-  const func = object[key];
-  if (! isManaged(func)) {
-    (object[key] = function wrapperFunc(param, filename) {
+function ensureWrapperMap(object, key) {
+  let map = getWrapperMap(object, key);
+  if (map === null) {
+    const func = object[key];
+    map = createWrapperMap(func);
+
+    object[key] = function (param, filename) {
       const pkgInfo = utils.getPkgInfo(path.dirname(filename));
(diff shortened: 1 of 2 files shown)


jdalton:
You're right about the expectation. Having folks copy metadata would be unexpected.

node/wrappers.js
```diff
@@ -21,9 +21,10 @@ exports.addWrapper = addWrapper;
 function createWrapperManager(object, key) {
   const func = object[key];
   if (! isManaged(func)) {
-    (object[key] = function(param, filename) {
+    (object[key] = function wrapperFunc(param, filename) {
       const pkgInfo = utils.getPkgInfo(path.dirname(filename));
-      const wrapper = pkgInfo === null ? null : findWrapper(object[key], pkgInfo.range);
+      const wrapper = pkgInfo === null ? null :
+        findWrapper(wrapperFunc, pkgInfo.range);
```
jdalton: The problem with using `wrapperFunc` as the anchor is once the `_extensions[".js"]` is wrapped by another we'll stop getting new entries added to `wrapperFunc` and they'll be added to the new wrapped func as we'll detect that it's no longer managed by us.

Instead we could use a symbol on the `_extensions` object to store the metadata which should persist through the hooks, `'.js'` and friends, being wrapped and rewrapped.

benjamn: Ok great, I'll update to that strategy.

node/wrappers.js
```diff
@@ -7,34 +7,60 @@ const isObject = require("../lib/utils.js").isObject;
...
 const reifyVersion = reifySemVer.version;
+const hasOwn = Object.prototype.hasOwnProperty;
 
-function addWrapper(func, wrapper) {
-  const reified = func.reified;
-  if (typeof reified.wrappers[reifyVersion] !== "function") {
-    reified.versions.push(reifyVersion);
-    reified.wrappers[reifyVersion] = wrapper;
+function getSymbolForKey(key) {
+  return Symbol.for("reifyWrapperMap$" + key);
```
benjamn: Bikeshed opportunity here, since we probably can't ever change the way we create this `Symbol`, without breaking backwards compatibility.

jdalton: We could make use of a generic `Symbol('__reify')` for entries and this metadata. For this case we can just the `Symbol('__reify')` to hold an object instead of using a symbol per key. That way we try to keep symbol namespace to a minimum and its something unlikely to change.

jdalton: Don't worry about it for this PR though.
I'll do a cleanup pass and tie everything together :bowtie:

jdalton:
Noice!

PR #159, opened 2017-05-08
Use module.watch(require(id), ...) to avoid calling module.require.

benjamn:
If we choose to go in this direction, we would be removing a significant barrier to interop with CommonJS module systems other than Node (in particular, Webpack and Browserify), since `module.require` would no longer be a requirement, and third-party bundling tools could easily understand dependencies embedded in `module.watch(require(id), ...)` calls.

I'm not committed to module.watch as a method name, though it's short and accurately reflects that the parent module is watching an exports object returned by a child module.

### DIFF
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -183,16 +183,18 @@ class ImportExportVisitor extends Visitor {
 
   visitExportAllDeclaration(path) {
     const decl = path.getValue();
-    const hoistedCode =
-      pad(this, "module.importSync(", decl.start, decl.source.start) +
-      getSourceString(this, decl) +
-      pad(
-        this,
-        ',{"*":function(v,k){exports[k]=v}},' +
-          makeUniqueKey(this) + ");",
-        decl.source.end,
-        decl.end
-      );
+    const hoistedCode = pad(
+      this,
+      "module.watch(require(" + getSourceString(this, decl) + ")",
+      decl.start,
+      decl.source.start
+    ) + pad(
+      this,
+      ',{"*":function(v,k){exports[k]=v}},' +
+        makeUniqueKey(this) + ");",
+      decl.source.end,
+      decl.end
+    );
 
     hoistExports(this, path, hoistedCode);
   }
@@ -726,7 +728,7 @@ function safeParam(param, locals) {
 }
 
 function toModuleImport(source, specifierMap, uniqueKey) {
-  let code = "module.importSync(" + source;
+  let code = "module.watch(require(" + source + ")";
   const importedNames = Object.keys(specifierMap);
   const nameCount = importedNames.length;
 
--- a/lib/runtime.js
+++ b/lib/runtime.js
@@ -11,6 +11,7 @@ exports.enable = function (mod) {
       typeof mod.importSync !== "function") {
     mod.export = moduleExport;
     mod.importSync = importSync;
+    mod.watch = moduleWatch;
     mod.runModuleSetters = runModuleSetters;
     return true;
   }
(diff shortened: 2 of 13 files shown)


benjamn:
A really interesting question is whether a package that uses a version of Reich that generates `module.importSync` calls could work alongside a package that generates `module.watch` calls. I think probably yes, but it would be nice to verify.

lib/runtime.js
```diff
@@ -11,25 +11,28 @@ exports.enable = function (mod) {
...
 // inside loops. The compiler generates these keys automatically (and
 // deterministically) when compiling nested import declarations.
 function importSync(id, setters, key) {
-  utils.setESModule(this.exports);
-  var exported = this.require(id);
-
-  if (utils.isObject(setters)) {
-    Entry.getOrCreate(exported).addSetters(this, setters, key);
-  }
+  return moduleWatch.call(this, this.require(id), setters, key);
```
jdalton: `this.watch(this.require(id), setters, key)`

jdalton:
Would this deprecate `importSync`, with the plan of removing it in a near future version?

jdalton:
*heads up I rebased this PR.

jdalton:
After this is merged it may be a good time to revisit the problem of referencing `module` and `require` in the generated code (since user code could redefine them). I'm cool just creating uniqish vars and using them.

benjamn:
> Would this deprecate `importSync`, with the plan of removing it in a near future version?

Yes, I think `importSync` will be unnecessary in the future, and we should remove it, though I vaguely worry about breaking existing generated code by changing the runtime.

jdalton:
Cool. We could deprecate it in v0.11 and remove it in v0.12 (that's being super nice)

jdalton:
By creating uniqish vars I'm mean something as simple as `__reifyModule` and `__reifyRequire` :P

benjamn:
Do you have any preferred techniques for giving not-too-noisy deprecation warnings?

jdalton:
I'd put it in the readme for v0.11 and you can add a note on npm install with [npm deprecate](https://docs.npmjs.com/cli/deprecate)

lib/import-export-visitor.js
```diff
@@ -184,8 +184,8 @@ class ImportExportVisitor extends Visitor {
   visitExportAllDeclaration(path) {
     const decl = path.getValue();
     const hoistedCode =
-      pad(this, "module.importSync(", decl.start, decl.source.start) +
-      getSourceString(this, decl) +
+      pad(this, "module.watch(", decl.start, decl.source.start) +
+      "require(" + getSourceString(this, decl) + ")" +
```
jdalton: Should the `"require(" + getSourceString(this, decl) + ")"` bit be in either the first `pad()` or the following one?

benjamn: Good catch, I think it should go in the one above, since it's usually (always?) only one line, and that line is the first line of the `module.watch` expression.

benjamn:
> After this is merged it may be a good time to revisit the problem of referencing `module` and `require` in the generated code (since user code could redefine them). I'm cool just creating unique-ish vars and using them. By creating unique-ish vars I'm mean something as simple as `__reifyModule` and `__reifyRequire` 😋 

I definitely agree we should do that for `module`. I think `require` is much safer, though, since most people know they'll mess up their Webpack bundles if they rename/shadow the `require` function.

lib/import-export-visitor.js
benjamn: Fixed, thanks!

benjamn:
Note to self: update the `README.md` to reflect these changes.

PR #160, opened 2017-05-09
Rename module.runModuleSetters to just module.runSetters.

benjamn:
The redundancy of this method name has always bothered me, so I figured I should deal with it at the same time as the `module.importSync` → `module.watch` changeover (#159), since the risks are essentially the same.

Like `module.importSync`, `module.runModuleSetters` is still defined in `lib/runtime.js`, but it will likely be deprecated and removed in a future version of Reify.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,24 +202,23 @@ simply becomes
 exports.default = getDefault();
 ```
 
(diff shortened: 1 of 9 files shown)


jdalton:
Nice!

PR #161, opened 2017-05-12
Add support for import()

jdalton:
This PR adds support for `import()` calls.

### DIFF
--- a/test/dynamic-import-tests.js
+++ b/test/dynamic-import-tests.js
@@ -1,0 +1,10 @@
+const assert = require("assert");
+
+describe("dynamic import", () => {
+  it("should support the import() function", (done) => {
+    import("./misc/abc").then((ns) => {
+      assert.deepEqual(ns, { a: "a", b: "b", c: "c" });
+      done();
+    });
+  });
+});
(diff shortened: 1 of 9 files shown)


test/dynamic-import-tests.js
```diff
@@ -0,0 +1,61 @@
...
+    assert.ok(p() instanceof Promise);
+  });
+
+  it("should support import() with yeild", () => {
+    function* p() {
+      yield import("./misc/abc");
+    }
+
+    assert.ok(p());
+  });
```
benjamn: Can we add a test that the properties of the namespace object get updated when the exported values change?

jdalton: Yes!

jdalton: Added.

jdalton:
👏 👏 👏 🎬 🗜

test/dynamic-import-tests.js
```diff
@@ -0,0 +1,76 @@
...
+  it("should support a template string id", (done) => {
+    const id = "./misc/abc";
+    import(`${id}`).then((ns) => {
+      assert.deepEqual(ns, { a: "a", b: "b", c: "c" });
+      done();
+    });
+  });
+
+  it("should establish live binding of values", (done) => {
+    import("./misc/live").then((ns) => {
```
benjamn: For future reference, Mocha lets you return a `Promise` from the test function rather than having to call the `done` function.

PR #162, opened 2017-05-13
Fix ns export when between a default export and exported specifiers.

jdalton:
Hey @benjamn. I found this bug but the \babylon compiler. Its having problems with:
```
export def4, * as ns4, { c as e } from "../misc/abc";
```
The acorn parser works because of our extension (bug fix).

I've skipped the from-extensions from other babylon specific tests, but it will still error when using the babylon parser (the first iteration of the run.sh).

### DIFF
--- a/lib/parsers/acorn-extensions/export.js
+++ b/lib/parsers/acorn-extensions/export.js
@@ -4,15 +4,8 @@ const acorn = require("acorn");
 const tt = acorn.tokTypes;
 
 exports.enable = function (parser) {
-  parser.checkExports = checkExports;
   parser.parseExport = parseExport;
 };
-
-function checkExports(exported, name) {
-  if (exported !== void 0) {
-    exported[name] = true;
-  }
-}
 
 function parseExport(node, exported) {
   // export ...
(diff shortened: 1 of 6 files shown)


jdalton:
They're having problems with:
```
export def4, * as ns4, { c as e } from "../misc/abc";
```
The acorn parser works because of our extension (bug fix).

jdalton:
Should I filter out babylon and plugin tests (as other tests do) or is this something that needs to be addressed in the babel plugin?

lib/parsers/acorn-extensions/export.js
```diff
@@ -156,10 +160,6 @@ function parseExportSpecifiersMaybe(parser, node) {
   }
 }
 
-function peekNextType(parser) {
-  return peekNextWith(parser, () => parser.type);
-}
```
jdalton: The bug was we were passing `() => parser.type` instead of `(p) => p.type`.

jdalton:
It turns out by spec order is significant, but I think that's goofy. I'll pull babylon until a proposal lands to support unordered import/export.

PR #163, opened 2017-05-13
Wrap module bodies with module.run(function(){...}).

benjamn:
This solves several subtle problems:

1. It ensures modules run with an undefined `this` binding.
2. It ensures we call `module.runSetters()` after any compiled module finishes loading, even if we are not able to patch runtime methods like `Module.prototype.{_compile,load}` (e.g. in Webpack).
3. It allows setting `module.loaded` even in environments where that wouldn't normally happen, so that we can rely on it elsewhere.

This PR should not be merged yet, as it still needs tests.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -66,6 +66,13 @@ exports.compile = function (code, options) {
     });
 
     importExportVisitor.finalizeHoisting();
+
+    if (getOption(options, "wrapWithRun")) {
+      result.code = "module.run(function(){" +
+        magicString.toString() + "//*/\n});";
+
+      return result;
+    }
   }
 
   result.code = magicString.toString();
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -67,7 +67,7 @@ exports.compile = function (code, options) {
 
     importExportVisitor.finalizeHoisting();
 
-    if (getOption(options, "wrapWithRun")) {
+    if (! getOption(options, "repl")) {
       result.code = "module.run(function(){" +
         magicString.toString() + "//*/\n});";
 
--- a/lib/options.js
+++ b/lib/options.js
@@ -7,7 +7,11 @@ const defaultOptions = {
   // If not false, "use strict" will be added to any modules with at least
   // one import or export declaration.
   enforceStrictMode: true,
-  wrapWithRun: true,
+  // If true, generate code appropriate for an interactive REPL session.
+  // In particular, in the REPL we should not wrap individual commands
+  // with module.run(function(){...}), and options.repl also implies the
+  // falsity of options.generateLetDeclarations (if unspecified).
+  repl: false,
   generateArrowFunctions: true,
   generateLetDeclarations: false,
   sourceType: "unambiguous",
@@ -18,8 +22,21 @@ const defaultOptions = {
 };
 
 function get(options, name) {
-  const result = hasOwn.call(options, name) ? options[name] : void 0;
-  return result === void 0 ? defaultOptions[name] : result;
+  if (hasOwn.call(options, name)) {
+    const result = options[name];
+    if (result !== void 0) {
+      return result;
+    }
+  }
+
+  if (name === "generateLetDeclarations" &&
+      hasOwn.call(options, "repl")) {
+    // If options.generateLetDeclarations is unspecified and options.repl
+    // is specified, return true iff options.repl is falsy.
+    return ! options.repl;
+  }
+
+  return defaultOptions[name];
 }
 
 exports.get = get;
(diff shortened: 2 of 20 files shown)


lib/compiler.js
```diff
@@ -65,6 +65,13 @@ exports.compile = function (code, options) {
     });
 
     importExportVisitor.finalizeHoisting();
+
+    if (getOption(options, "wrapWithRun")) {
+      result.code = "module.run(function(){" +
+        magicString.toString() + "//*/\n});";
+
+      return result;
+    }
```
jdalton: Without `wrapWithRun` things break so doesn't seem like much of an option. Let's just use `module.run` and roll with it.

jdalton: I noticed you toggle `wrapWithRun` for the REPL but even the REPL should have a `this` of `undefined`. Does `run` cause an issue in the REPL?

benjamn: I think `this` should be whatever it is in the REPL (if you type `this` and hit return), so forcing it to be undefined seems undesirable there.

benjamn: > I'm down for just using `module.run` and rolling with it.

I agree, I just need to update a lot of tests after switching the default.

jdalton: > I think this should be whatever it is in the REPL (if you type this and hit return), so forcing it to be undefined seems undesirable there.

By opting into reify you're enabling esm in the repl so the executed code should be esm. Switching parse goals depending on code in the repl after opting in seems a bit odd.

benjamn: I'm not sure I agree that the REPL should be treated as module code. In particular, I think `this` should still refer to the global object, and the REPL should not force strict mode. Remember the REPL is already a bit weird because we have to generate `var` declarations rather than `let`. I think the goal should be simply to make `import` declarations work in the REPL, with as few other changes as possible.

benjamn: Also there's no point in running `module.runSetters()` for the REPL `module`, since no other code will be importing anything from the REPL.

jdalton: >Also there's no point in running module.runSetters() for the REPL module, since no other code will be importing anything from the REPL.

Ya, but `module.runSetters` doesn't hurt and toggling for repl ends up complicating several things for something that's a secondary, bonus, scenario. So what I'm saying is the REPL isn't worth the juggle cost. If it works great, but complicating the compiler, options, caching compiler, for it is a bit much.

jdalton: The only reason `wrapWithRun` exists is because of the REPL so I don't think the option should exist. Maybe instead `run` can detect it's in a REPL or maybe the repl could add its own `run` method that doesn't set the `this`.

benjamn: I'm pretty sure running all REPL code inside a function wrapper will cause more problems than it solves, and lead to unpleasant surprises.

For example, if I flip `wrapWithRun` to `true` in the REPL, I can't use any of the symbols I've imported, because the variable declaration ends up inside the `module.run` callback. Hoisting those variables outside the `module.run` callback feels like it would be a very big departure from the way things currently work (where the variable declaration is always right before the import).

benjamn: > The only reason `wrapWithRun` exists is because of the REPL so I don't think the option should exist.

Perhaps the compiler should be generally aware of whether it's compiling for the REPL, and do the right thing here (and in other cases). I agree the `wrapWithRun` option smells like an implementation detail, now that we're only toggling it in the REPL, but I'm still pretty convinced we should not wrap REPL code with `module.run(function(){...})`.

jdalton: >For example, if I flip wrapWithRun to true in the REPL, I can't use any of the symbols I've imported, because the variable declaration ends up inside the module.run callback. 

Ah ok, how does the normal module wrapper not interfere?

> If we had something like options.repl, we might even be able to get rid of options.generateLetDeclarations since the REPL is the only consumer that really needs to generate var declarations (unless I'm mistaken).

Oh niice! `options.repl` is nice and clean. We also have a `utils.isREPL()` helper to auto detect.

jdalton:
Could this be changed to the `options.repl` approach with the auto detect if not specified?

benjamn:
Yep, I'll switch to `options.repl` as we discussed.

benjamn:
Ok, the option is now called `options.repl` (with the opposite consequences of `options.wrapWithRun`), and we auto-detect it [here](https://github.com/benjamn/reify/blob/0c9cbfe9ed6cbab57b1a670aacf3011e0e0eee02/node/repl-hook.js#L28).

jdalton:
Let's do it!

jdalton:
Oh crap, lol I just read about the tests. Well let's add tests now . No time like the present 😃

PR #164, opened 2017-05-13
Keep unique reference to `module` to avoid conflict with user variables.

### LINKED ISSUE #139: Use a unique reference for `module` in case the user creates their own local identifier.

benjamn:
The `makeUniqueIdQuickly` function scans the source of the module for anything that looks like a `/\bmodule\d*\b/` identifier, then returns an identifier name starting with `module` and ending with a numeric suffix (if necessary) that is guaranteed not to exist anywhere in the source.

Fixes #139.

### DIFF
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -49,8 +49,10 @@ exports.compile = function (code, options) {
     return result;
   }
 
-  const rootPath = new FastPath(ast);
+  options.moduleId = options.moduleId ||
+    makeUniqueIdQuickly("module", code);
 
+  const rootPath = new FastPath(ast);
   importExportVisitor.visit(rootPath, code, options);
   result.identical = ! importExportVisitor.madeChanges;
 
@@ -67,9 +69,27 @@ exports.compile = function (code, options) {
   }
 
   result.code = magicString.toString();
+
   return result;
 };
 
 exports.transform = function (ast, options) {
   return dynRequire("./transform.js")(ast, options);
 };
+
+function makeUniqueIdQuickly(prefix, source) {
+  const scanRegExp = new RegExp("\\b" + prefix + "(\\d*)\\b", "g");
+  let match, max = -1;
+
+  while ((match = scanRegExp.exec(source))) {
+    max = Math.max(max, +(match[1] || 0));
+  }
+
+  if (max >= 0) {
+    return prefix + (max + 1);
+  }
+
+  return prefix;
+}
+
+exports.makeUniqueId = makeUniqueIdQuickly;
--- a/lib/compiler.js
+++ b/lib/compiler.js
@@ -50,7 +50,7 @@ exports.compile = function (code, options) {
   }
 
   options.moduleAlias = options.moduleAlias ||
-    makeUniqueIdQuickly("module", code);
+    makeUniqueId("module", code);
 
   const rootPath = new FastPath(ast);
   importExportVisitor.visit(rootPath, code, options);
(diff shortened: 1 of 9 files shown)


test/transform-tests.js
```diff
@@ -15,6 +19,8 @@ describe("compiler.transform", () => {
       const code = files[absPath];
       const ast = options.parse(code);
 
+      options.moduleId = makeUniqueId("module", code);
```
benjamn: It's a little unfortunate that `transform` can't come up with a unique `module` name on its own, but that's ok because only the Babel plugin currently uses `transform`, and the Babel plugin sets `options.moduleId` explicitly.

packages/babel-plugin-transform-es2015-modules-reify/index.js
```diff
@@ -74,7 +74,16 @@ module.exports = function () {
           parse: parse
         };
 
-        transform(path.node, Object.assign(transformOptions, this.opts));
+        var code = path.hub.file.code;
+        if (typeof code === "string") {
+          transformOptions.moduleId =
+            compiler.makeUniqueId("module", code);
```
benjamn: Here's where the Babel plugin sets `transformOptions.moduleId` explicitly.

lib/compiler.js
```diff
@@ -50,8 +50,10 @@ exports.compile = function (code, options) {
     return result;
   }
 
-  const rootPath = FastPath.from(ast);
+  options.moduleId = options.moduleId ||
+    makeUniqueIdQuickly("module", code);
```
benjamn: It's possible to pass in `options.moduleId` explicitly, to override/skip calling `makeUniqueIdQuickly`.

test/compiler-tests.js
```diff
@@ -34,6 +34,12 @@ describe("compiler", () => {
     assert.ok(/unexpected/i.test(error.message));
   });
 
+  it("should choose a unique module identifier", () => {
+    const module = null, module2 = null;
+    import { a } from "./misc/abc";
+    assert.strictEqual(a, "a");
+  });
```
benjamn: We should definitely have more tests than this.

lib/compiler.js
```diff
@@ -68,9 +70,27 @@ exports.compile = function (code, options) {
...
+    max = Math.max(max, +(match[1] || 0));
+  }
+
+  if (max >= 0) {
+    return prefix + (max + 1);
+  }
+
+  return prefix;
+}
+
```
jdalton: It can just be `makeUniqueId` like the exported name for it. Quickly is subjective n all.

lib/compiler.js
```diff
@@ -68,9 +70,27 @@ exports.compile = function (code, options) {
...
   return result;
 };
 
 exports.transform = function (ast, options) {
   return dynRequire("./transform.js")(ast, options);
 };
+
+function makeUniqueIdQuickly(prefix, source) {
+  const scanRegExp = new RegExp("\\b" + prefix + "(\\d*)\\b", "g");
```
jdalton: No need to tweak anything yet, but I'll look at benchmarking this vs using a canned regexp for the common case of "module".

test/transform-tests.js
jdalton: Bikeshed, `moduleId` is kind of a odd name. When I think of module id I think of the identifier of a module. Like `reify` is the module id of `require('reify')`. Maybe `moduleAlias`?

lib/compiler.js
jdalton: Besides the previous notes everything else is 👍

benjamn: I'll rebase this tomorrow morning and merge. Thanks for the review!

Issue #134, opened 2017-04-13
Don't save setter for const exports.


jdalton:
This would address a line item on #90. It would also be a good time to think about the parser optimization to do a top-level only parse if there are only const exports and nested imports is disabled.

benjamn:
Specifically, we can discard the setter after the first time it is called, if `module.loaded` is true, meaning we should have finished initializing all top-level `const` values.

PR #166, opened 2017-05-18
Simplify module.export in order to implement #134

benjamn:
This pull request
* compiles `export default <expr>` to `module.exportDefault(<expr>)` instead of `module.export("default", exports.default = <expr>)`, which means `module.export` no longer has to handle completely different argument types;
* signals when the values returned by getter functions will never change by calling `module.export(getters, true)`; and
* discards setter functions that will never be called again, fixing #134.

### DIFF
--- a/lib/entry.js
+++ b/lib/entry.js
@@ -67,7 +67,7 @@ Entry.getOrCreate = function (exported) {
   return entry;
 };
 
-Ep.addGetters = function (getters) {
+Ep.addGetters = function (getters, constant) {
   var names = Object.keys(getters);
   var nameCount = names.length;
 
@@ -81,6 +81,7 @@ Ep.addGetters = function (getters) {
         // Should this throw if this.getters[name] exists?
         ! (name in this.getters)) {
       this.getters[name] = getter;
+      getter.constant = !! constant;
     }
   }
 };
--- a/lib/entry.js
+++ b/lib/entry.js
@@ -70,6 +70,7 @@ Entry.getOrCreate = function (exported) {
 Ep.addGetters = function (getters, constant) {
   var names = Object.keys(getters);
   var nameCount = names.length;
+  constant = !! constant;
 
   for (var i = 0; i < nameCount; ++i) {
     var name = names[i];
@@ -81,7 +82,7 @@ Ep.addGetters = function (getters, constant) {
         // Should this throw if this.getters[name] exists?
         ! (name in this.getters)) {
       this.getters[name] = getter;
-      getter.constant = !! constant;
+      getter.constant = constant;
       getter.runCount = 0;
     }
   }
--- a/lib/entry.js
+++ b/lib/entry.js
@@ -82,6 +82,7 @@ Ep.addGetters = function (getters, constant) {
         ! (name in this.getters)) {
       this.getters[name] = getter;
       getter.constant = !! constant;
+      getter.runCount = 0;
     }
   }
 };
@@ -258,6 +259,23 @@ function forEachSetter(entry, names, callback) {
 
       } else {
         call(setters[key], name, value, callback);
+
+        var getter = entry.getters[name];
+        if (typeof getter === "function" &&
+            // Sometimes a getter function will throw because it's called
+            // before the variable it's supposed to return has been
+            // initialized, so we need to know that the getter function
+            // has run to completion at least once.
+            getter.runCount > 0 &&
+            getter.constant) {
+          // If we happen to know that this getter function has run
+          // successfully, and will never return a different value, then
+          // we can forget the corresponding setter, because we've already
+          // reported that constant value. Note that we can't forget the
+          // getter, because we need to remember the original value in
+          // case anyone tampers with entry.exports[name].
+          delete setters[key];
+        }
       }
     }
   }
@@ -290,8 +308,11 @@ function makeUniqueKey() {
 }
 
 function runGetter(entry, name) {
+  var getter = entry.getters[name];
   try {
-    return entry.getters[name]();
+    var result = getter();
+    ++getter.runCount;
+    return result;
   } catch (e) {}
   return GETTER_ERROR;
 }
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -26,14 +26,15 @@ class ImportExportVisitor extends Visitor {
         codeToInsert += '"use strict";';
       }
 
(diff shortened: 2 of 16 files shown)


test/export/later.js
```diff
@@ -4,15 +4,18 @@ export default "default-1";
...
 
 export function exportAgain() {
-  module.export("default", exports.default = "default-2");
+  // Neither of these re-export styles should work, because the original
+  // export default still takes precedence over anything else.
+  module.exportDefault(exports.default = "default-2");
+
+  // This style also does not work, because the getter function for the
+  // variable val is all that matters.
+  exports.val = "ignored";
```
benjamn: @jdalton Here's proof that tampering with `exports` directly does not interfere with `module.export` getters.

test/output/export-all/expected.js
```diff
@@ -1,2 +1,2 @@
 "use strict";module.watch(require("./abc"),{"*":function(v,k){exports[k]=v}},0);
-module.export("default",exports.default=("default"));
+module.exportDefault("default");
```
benjamn: Much cleaner!

test/setter/grandchild.js
```diff
@@ -1,5 +1,5 @@
 export let c = 0;
 export function increment() {
   ++c;
-  module.export();
+  module.runSetters();
```
benjamn: This `module.export()` call was a very old way of triggering setters, even less common than `module.export("default", ...)`. Glad to see it gone!

lib/entry.js
```diff
@@ -81,6 +81,8 @@ Ep.addGetters = function (getters) {
         // Should this throw if this.getters[name] exists?
         ! (name in this.getters)) {
       this.getters[name] = getter;
+      getter.constant = !! constant;
```
jdalton: `constant` can be coerced to a boolean outside the loop (move to top)

lib/entry.js
```diff
@@ -257,6 +259,23 @@ function forEachSetter(entry, names, callback) {
 
       } else {
         call(setters[key], name, value, callback);
+
+        var getter = entry.getters[name];
+        if (typeof getter === "function" &&
+            // Sometimes a getter function will throw because it's called
+            // before the variable it's supposed to return has been
+            // initialized, so we need to know that the getter function
+            // has run to completion at least once.
```
jdalton: Do we have a test for this throw case?

lib/import-export-visitor.js
```diff
@@ -225,15 +224,28 @@ class ImportExportVisitor extends Visitor {
...
+      if (dd.type === "SequenceExpression") {
+        // If the exported expression is a comma-separated sequence
+        // expression, this.code.slice(dd.start, dd.end) may not include
+        // the vital parentheses, so we should wrap the expression with
+        // parentheses to make absolutely sure it is treated as a single
+        // argument to the module.exportDefault method, rather than as
+        // multiple arguments.
+        prefix += "(";
+        suffix = ")" + suffix;
+      }
```
jdalton: This sequence check looks new. Do we have a test for it?

lib/import-export-visitor.js
```diff
@@ -245,7 +257,22 @@ class ImportExportVisitor extends Visitor {
           dd.type = "ClassExpression";
         }
 
-        path.replace(buildExportDefaultStatement(this, dd));
+        // Almost every JS parser parses this expression the same way, but
+        // we should still give custom parsers a chance to parse it.
+        let ast = this.parse("module.exportDefault(ARG);");
```
jdalton: If you're giving custom parsers a chance should you then also be enforcing its structure with asserts below?

lib/import-export-visitor.js
```diff
@@ -598,6 +613,25 @@ function hoistExports(visitor, exportDeclPath, mapOrString, childName) {
   visitor.madeChanges = true;
 }
 
+function isConstExportDeclaration(exportDecl) {
```
jdalton: This is really about live binding not const though right? The export default declaration isn't a const so the name is a bit confusing. Maybe `isLiveBound`?

test/output/export-all/expected.js
jdalton: Nice!

lib/import-export-visitor.js
benjamn: We [do](https://github.com/benjamn/reify/tree/b84bd5695a34059b1580281e29b5c610a0ed3e79/test/output/default-expression)! Previously, we were just wrapping the argument with parentheses unconditionally, but now we try to avoid it except in this one very strange case.

lib/import-export-visitor.js
benjamn: I'm enforcing the things I care about, but the custom parser might have its own metadata that I don't care about.

lib/import-export-visitor.js
jdalton: Oh nice!

lib/import-export-visitor.js
benjamn: I see what you're saying, but I think maybe `canExportedValuesChange` would be an even more accurate name, since "live bound" refers to the native inter-module scope-sharing 
functionality that we're trying to simulate.

lib/entry.js
benjamn: I'll write one.

lib/import-export-visitor.js
jdalton: I can dig!

lib/entry.js
benjamn: Added a test that I think demonstrates this `getter.runCount > 0` check matters, even though the test itself is a bit artificial.

test/setter-tests.js
```diff
@@ -57,4 +57,41 @@ describe("parent setters", () => {
...
+    // value here is undefined, and our live binding to that value is
+    // maintained for now.
+    assert.strictEqual(value, void 0);
+    assert.strictEqual(valueSetterCallCount, 1);
+
+    // Since the exported value is marked as constant, the setter that
+    // updates our copy of value should be discarded as soon as it has
+    // been called with the final value, but this set(2) still works
+    // because the getter threw the last time it was called. After this,
+    // however, the setter function will be discarded.
```
benjamn: Although this whole test is a bit artificial, this logic breaks if we remove the `getter.runCount > 0` test in `lib/entry.js`, because we forget the setter too soon.

jdalton:
:shipit:

PR #170, opened 2017-05-22
Add failing test due to missing module.runSetters() call.

benjamn:
This test fails because module.runSetters() is not called when the `test/setter/cycle/cjs.js` module finishes evaluating, due to this commit: 7932aa56cef2ead61084c3b67a5afa329fe5415e

The general problem is that we need to call `module.runSetters()` even for modules that were not compiled by Reify, or wrapped with `module.run(...)`, so that other modules can import live bindings from them.

If you add `module.runSetters()` to the end of cjs.js, or add back the `module.runSetters()` call in the compile hook, this test will pass.

### DIFF
--- a/node/compile-hook.js
+++ b/node/compile-hook.js
@@ -57,6 +57,11 @@ function extWrap(func, pkgInfo, mod, filePath) {
 
   runtime.enable(mod);
   mod._compile(cacheValue, filePath);
+
+  // If the module is not loaded through module.run, then run its setters.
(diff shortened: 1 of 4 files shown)


jdalton:
Neat!

jdalton:
Patched the compile-hook to run setters if not loaded through `module.run`. It uses the `loaded` flag as an indicator since `module.run` will set it to `true` otherwise it would be `false` at this point.

PR #172, opened 2017-05-22
Pass namespaces to module.watch so child module can manage them.

### LINKED ISSUE #171: Namespace objects should come from child module, not parent
After [this recent commit](https://github.com/benjamn/reify/commit/39be6fe29ed1e738c6c31a33204792bc1814bb04), the `module.run` API now `Object.seal`s all namespace objects created by the module when the module is finished evaluating.

The problem is, when we create namespace objects, it's the parent module that ends up tracking them, when what we really want is for the child/imported module to be the one that creates and later seals the namespace objects.

For example, here's what I think needs to change in the `module.import` API:

```diff
commit 162418d52dc51f4311b1011524f1a91ad990e2bf
Author: Ben Newman <ben@benjamn.com>
Date:   Mon May 22 12:28:59 2017 -0400

    Use childEntry.createNs to create namepace objects.

diff --git a/lib/runtime/entry.js b/lib/runtime/entry.js
index 05683f9..05cc0e0 100644
--- a/lib/runtime/entry.js
+++ b/lib/runtime/entry.js
@@ -97,11 +97,22 @@ function addGetters(getters, constant) {
 
 Ep.addGetters = addGetters;
 
-function addNs(ns) {
+Ep.createNs = function () {
+  var ns = Object.create(null);
+
+  if (useToStringTag) {
+    Object.defineProperty(ns, Symbol.toStringTag, {
+      configurable: false,
+      enumerable: false,
+      value: "Module",
+      writable: false
+    });
+  }
+
   this.nsObjects.push(ns);
-}
 
-Ep.addNs = addNs;
+  return ns;
+};
 
 function addSetters(parent, setters, key) {
   var names = Object.keys(setters);
diff --git a/lib/runtime/index.js b/lib/runtime/index.js
index 6ebc7a5..1cba4a4 100644
--- a/lib/runtime/index.js
+++ b/lib/runtime/index.js
@@ -33,19 +33,7 @@ function enable(mod) {
 Runtime.enable = enable;
 
 function createNs() {
-  var ns = Object.create(null);
-
-  if (useToStringTag) {
-    Object.defineProperty(ns, Symbol.toStringTag, {
-      configurable: false,
-      enumerable: false,
-      value: "Module",
-      writable: false
-    });
-  }
-
-  Entry.getOrCreate(this.exports).addNs(ns);
-  return ns;
+  return Entry.getOrCreate(this.exports).createNs();
 }
 
 Rp.createNs = createNs;
@@ -84,13 +72,16 @@ function moduleImport(id) {
   var that = this;
 
   return Promise.resolve().then(function () {
-    var ns = that.createNs();
-    that.watch(that.require(id), {
+    var exported = that.require(id);
+    var ns = Entry.getOrCreate(exported).createNs();
+
+    that.watch(exported, {
       "*": function (value, name) {
         ns[name] = value;
       }
     });
-    return finishNs(ns);
+
+    return ns;
   });
 }
```

However, we also generate code that calls `module.createNs()` on the wrong (parent) module, and I'm less sure how we can fix that:
```js
import * as ns from "./child";
```
becomes
```js
const ns = module.createNs();
module.watch(require("./child"), {
  "*": (value, name) => {
    ns[name] = value;
  }
});
```
Somehow we have to let `createNs` know what child module will be managing the namespace, so perhaps we could pass something to `createNs`, like the child's `exports` object?
```js
const ns = module.createNs(require("./child"));
module.watch(require("./child"), {
  "*": (value, name) => {
    ns[name] = value;
  }
});
```
That would be enough for `createNs` to record the resulting namespace object properly with the child module, though I don't love calling `require("./child")` twice.

Any ideas?

benjamn:
Fixes #171.

With these changes, an `import` declaration like
```js
import * as ns from "./child";
```
gets transformed to
```js
const ns = Object.create(null);
module.watch(require("./child"), {
  "*": (value, name) => {
    ns[name] = value;
  }
}, key, [ns]);
```
and an `export` declaration like
```js
export * as ns from "./child";
```
gets transformed to
```js
exports.ns = Object.create(null);
module.watch(require("./child"), {
  "*": (value, name) => {
    exports.ns[name] = value;
  }
}, key, [exports.ns]);
```
The `module.watch` API is now responsible for setting up the namespace object and remembering it until it can be sealed, because no new properties need to be added.

### DIFF
--- a/lib/runtime/entry.js
+++ b/lib/runtime/entry.js
@@ -21,9 +21,11 @@ var entryKey = useSymbol ? entrySymKey : entryStrKey;
 var hasOwn = Object.prototype.hasOwnProperty;
 var keySalt = 0;
 
-function Entry(exported) {
+function Entry(exported, owner) {
   // The module.exports of the module this Entry is managing.
   this.exports = exported;
+  // The module object this Entry is managing, if known.
+  this.owner = owner || null;
   // Getters for local variables exported from the managed module.
   this.getters = Object.create(null);
   // Namespace objects created in the managed module.
@@ -47,20 +49,20 @@ function get(exported) {
 
 Entry.get = get;
 
-function getOrCreate(exported) {
+function getOrCreate(exported, owner) {
   if (! utils.isObjectLike(exported)) {
     // In case the child module modified module.exports, create a temporary
     // Entry object so that we can call the entry.addSetters method once,
     // which will trigger entry.runSetters(names), so that module.importSync
     // behaves as expected.
-    return new Entry(exported);
+    return new Entry(exported, owner);
   }
 
   if (hasOwn.call(exported, entryKey)) {
     return exported[entryKey];
   }
 
-  var entry = new Entry(exported);
+  var entry = new Entry(exported, owner);
 
   if (useSymbol) {
     exported[entryKey] = entry;
--- a/lib/runtime/entry.js
+++ b/lib/runtime/entry.js
@@ -153,7 +153,7 @@ function addSetters(parent, setters, key, namespaces) {
 
   this.runSetters(names);
 
-  if (utils.isArray(namespaces)) {
+  if (Array.isArray(namespaces)) {
     // If any namespace objects were provided, remember them so that we
     // can call Object.seal(namespace) later, when the module this Entry
     // is managing has finished loading.
--- a/lib/runtime/entry.js
+++ b/lib/runtime/entry.js
@@ -12,6 +12,7 @@ var NAN = {};
 var UNDEFINED = {};
 
 var useSymbol = typeof Symbol === "function";
+var useToStringTag = useSymbol && typeof Symbol.toStringTag === "symbol";
 
 var entryStrKey = "__reifyEntry";
 var entrySymKey = useSymbol ? Symbol.for(entryStrKey) : null;
@@ -29,6 +30,9 @@ function Entry(exported) {
   this.nsObjects = [];
   // Setters for assigning to local variables in parent modules.
   this.setters = Object.create(null);
+  // Boolean indicating whether the module this Entry is managing has
+  // finished evaluation yet.
+  this.loaded = false;
 }
 
 var Ep = utils.setPrototypeOf(Entry.prototype, null);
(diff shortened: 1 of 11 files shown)


benjamn:
Are there actually any cases where we would have more than one namespace object per `module.watch` call? Does `export * as a, * as b from "./child"` work?

jdalton:
Everything besides the unneeded `Array.isArray` polyfill looks great. All my 👍 are belong to this.

lib/runtime/entry.js
```diff
@@ -29,6 +30,9 @@ function Entry(exported) {
   this.nsObjects = [];
   // Setters for assigning to local variables in parent modules.
   this.setters = Object.create(null);
+  // Boolean indicating whether the module this Entry is managing has
+  // finished evaluation yet.
+  this.loaded = false;
```
jdalton: All like all the `loaded` state stuff. :yum

node/compile-hook.js
```diff
@@ -60,6 +60,7 @@ function extWrap(func, pkgInfo, mod, filePath) {
 
   // If the module is not loaded through module.run, then run its setters.
   if (! mod.loaded) {
+    mod.loaded = true;
```
benjamn: This is necessary to trigger `entry.onLoaded()` in `mod.runSetters()`.

lib/runtime/entry.js
```diff
@@ -97,13 +101,26 @@ function addGetters(getters, constant) {
...
+
+  if (this.loaded) {
+    Object.seal(ns);
+  } else {
+    this.nsObjects.push(ns);
+  }
 }
 
-Ep.addNs = addNs;
+Ep.addNamespace = addNamespace;
```
jdalton: I like the more verbose name!

node/compile-hook.js
jdalton: I can dig the explicit `loaded` being set here. (It's set by `Module.prototype.load` after this call but this clears things up.)

jdalton: I can dig it. `loaded` looks like Lodash to me and so I say MOAR !

benjamn:
Thanks for the review!

I have one more concern to work out, and it has to do with `module.exports = require(...)` bridge modules. I'm worried the bridge module might "finish" loading before the required module is really finished loading, which would result in calling `Entry.getOrCreate(module.exports).onLoaded()` too early. Trying to think of a more reliable way to manage `entry.loaded`.

jdalton:
> Are there actually any cases where we would have more than one namespace object per `module.watch` call? Does `export * as a, * as b from "./child"` work?

Good Q. You can do this for sure:
```js
export * as b from "./child"
export * as a from "./child"
```

jdalton:
>  Does export * as a, * as b from "./child" work?

No, having more than 1 will throw an assert error in import-export-visitor.

PR #174, opened 2017-05-23
House cleaning for Entry additions.

jdalton:


lib/runtime/entry.js
```diff
@@ -115,9 +115,9 @@ function addNamespace(ns) {
 
   if (this.loaded) {
     Object.seal(ns);
-  } else {
-    this.nsObjects.push(ns);
   }
+
+  this.namespaces.push(ns);
```
jdalton: This also tracks the ns objects even after load in case we want to query them for some reason.

lib/runtime/entry.js
```diff
@@ -24,17 +24,17 @@ var keySalt = 0;
...
-  this.nsObjects = [];
-  // Setters for assigning to local variables in parent modules.
-  this.setters = Object.create(null);
   // Boolean indicating whether the module this Entry is managing has
   // finished evaluation yet.
   this.loaded = false;
+  // Namespace objects created in the managed module.
+  this.namespaces = [];
+  // The module object this Entry is managing, if known.
+  this.owner = utils.isObject(owner) ? owner : null;
```
jdalton: Uses an `isObject` check instead of a bool fall through.

lib/runtime/entry.js
benjamn: Ok!

lib/runtime/index.js
```diff
@@ -165,9 +165,9 @@ function watch(exported, setters, key, namespaces) {
 
 Rp.watch = watch;
 
-function setESModuleAndOwner(module) {
-  utils.setESModule(module.exports);
-  Entry.getOrCreate(module.exports, module);
```
jdalton: This helps prevent confusion since we know by looking at it that it's not the real real `module`.

PR #175, opened 2017-05-23
Add support for specifying multiple namespace exports in a clause.

jdalton:
Per our discussion this PR removes the restriction on export clauses so that more than 1 namespace export is allowed.
```js
export * as a, * as b from "./child"
```

### DIFF
--- a/lib/import-export-visitor.js
+++ b/lib/import-export-visitor.js
@@ -26,27 +26,16 @@ class ImportExportVisitor extends Visitor {
 
       if (bodyInfo.parent.type === "Program" &&
           this.moduleAlias !== "module") {
-        codeToInsert += this.generateLetDeclarations ? "const " : "var ";
-        codeToInsert += this.moduleAlias + "=module;";
+        codeToInsert +=
+          (this.generateLetDeclarations ? "const " : "var ") +
+          this.moduleAlias + "=module;";
       }
 
-      const addExportsMap = (map, constant) => {
-        const namedExports = toModuleExport(this, map, constant);
-        if (namedExports) {
-          codeToInsert += namedExports;
-        }
-      };
-
-      addExportsMap(bodyInfo.hoistedExportsMap, false);
-      addExportsMap(bodyInfo.hoistedConstExportsMap, true);
-
-      if (bodyInfo.hoistedExportsString) {
-        codeToInsert += bodyInfo.hoistedExportsString;
-      }
-
-      if (bodyInfo.hoistedImportsString) {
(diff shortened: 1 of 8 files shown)


lib/import-export-visitor.js
```diff
@@ -794,12 +779,17 @@ function toModuleImport(visitor, source, specifierMap, namespaces) {
   }
 
   const lastIndex = nameCount - 1;
+  let namespaceList = Array.isArray(namespaces) ? namespaces.join(",") : "";
+  const searchExports = ! namespaceList;
```
jdalton: This avoids modifying the provided `namespaces` array and uses the lack of an array as an indicator to search for `exports.`.

test/output/export-multi-namespace/expected.js
```diff
@@ -0,0 +1,3 @@
+module.run(function(){"use strict";var module1=module;exports.a=Object.create(null);exports.b=Object.create(null);module1.watch(require("module"),{"*":function(v,n){exports.a[n]=exports.b[n]=v}},0,[exports.a,exports.b]);
```
jdalton: As a bonus this now defines the specifiers in the correct order. Before I had them reversed.

test/output-tests.js
```diff
@@ -27,7 +27,10 @@ Object.keys(files).forEach((absPath) => {
 
 describe("output", () => {
   function check(data) {
-    const code = compile(data.actual).code;
+    const code = compile(data.actual, {
+      parse: require("../lib/parsers/acorn.js").parse
```
jdalton: Do you want to force all parser passes to use this or should we add in skip logic similar to other areas.

test/export/extensions.js
```diff
@@ -3,4 +3,4 @@ export { a, b as c }, * as ns2 from "../misc/abc";
 export def1 from "../misc/abc";
 export def2, { b, c as d } from "../misc/abc";
 export def3, * as ns3 from "../misc/abc";
-export * as ns4, { c as e }, def4 from "../misc/abc";
+export * as ns4, * as ns5, { c as e }, def4, def5 from "../misc/abc";
```
jdalton: FWIW this PR had a test ^

benjamn: I wanted to have an output test, too (this is just a behavior test).

test/output-tests.js
benjamn: I think Acorn is a better default here, since it's the default anyway, and we have more control over how it parses new syntax.

PR #176, opened 2017-05-23
Rename parsers/default.js to parsers/index.js.

jdalton:
Since this feels like a place to use an index.js this PR does just that.


benjamn:
Hmm, couple thoughts:
* The plurality of `require("reify/lib/parsers").parse` feels a bit weird, since you're requiring one parser.
* Some external code reaches into the `reify` package to use the same parser that it uses, so this would be a breaking change. But I guess that just means it needs to go into 0.12.0.

jdalton:
Ah the plurality bit. It could read well as `reify/lib/parser/acorn` so I can mod that.... then the only plural bit would be the `acorn-extensions` which could be singular too and read well.

lib/parser/acorn.js
```diff
@@ -1,7 +1,7 @@
 "use strict";
 
 const acorn = require("acorn");
-const acornExtensions = require("./acorn-extensions");
+const acornExtension = require("./acorn-extension");
```
benjamn: Sorry for the back-and-forth, but I think the plural form here is better!

jdalton: Np. Variety is the spice of life ✨

PR #177, opened 2017-05-24
Further harden REPL detect by checking module.children.

jdalton:
This takes the check a bit further and only enables the wrapper if reify was explicitly required from the REPL.

### DIFF
--- a/lib/env.js
+++ b/lib/env.js
@@ -1,0 +1,7 @@
+"use strict";
+
+module.exports = Object.assign(Object.create(null),
+  typeof process === "object" && process !== null &&
+  typeof process.env === "object" && process.env !== null &&
+  process.env
+);
--- a/README.md
+++ b/README.md
@@ -5,7 +5,7 @@
   1. to make (something abstract) more concrete or real<br>
      _"these instincts are, in humans, reified as verbal constructs"_
   2. to regard or treat (an idea, concept, etc.) as if having material existence
-  3. **to enable [ECMAScript 2015 modules](http://www.2ality.com/2014/09/es6-modules-final.html) in *any* version of [Node.js](https://nodejs.org)**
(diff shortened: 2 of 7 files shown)


lib/env.js
```diff
@@ -0,0 +1,7 @@
+"use strict";
+
+module.exports = (
+  typeof process === "object" && process !== null &&
+  typeof process.env === "object" && process.env !== null &&
+  process.env
+) || {};
```
jdalton: Is the `process` guard still needed here? I don' think it's because of PhantomJS.
Maybe some Meteor build thing?

benjamn:
It's a bit unfortunate that, in Node, `module.children` only contains child modules if `module` was the first module to import those children. In other words,
```js
parent.children.every(child => child.parent === parent)
```
must be true. This condition holds, for better or worse, because [`parent.children.push(this)`  is only called in the `Module` constructor](https://github.com/nodejs/node/blob/e912c67d24ba42eeb47431a2b163f7f8a8532c78/lib/module.js#L55), rather than whenever a child module is required.

This is also why I had to add [`module.childrenById`](https://github.com/benjamn/install/blob/fa332017a584b3b643ee999e44f5e6146e1bdba0/install.js#L82-L90) in the `install.js` implementation, so that I could actually rely on `module.children`-like information.

Does that gotcha have any consequences for this PR?

jdalton:
In the repl whenever you require a module it's added to its `.children`. But you're right if a child loads reify then it won't be added to the repls children when the repl loads reify at a later time. So I'll dig a bit more to resolve that.

jdalton:
Ok so thinking on this a bit more... I'm OK with saying that folks should load reify first thing to enable the abilities otherwise there's no way to determine if it's being loaded by a subsequent package. This likely aligns with the common usage too, "Pop in to the REPL, load up reify, and start using ESM".

jdalton:
Updated the doc note to call it out.

jdalton:
Let's hold off until there's a user reported issue.

PR #178, opened 2017-05-24
Move module.run wrapper to compile-hook.

jdalton:
This PR moves the `module.run` wrapper to the compile-hook. This allows us to customize wrapper per option/source-type and simplifies parsing a bit.

Things to note:

* The compiler `result` now includes a `sourceType` that specifies what type it compiled as

* The `enforceStrictMode` option is removed since `"use strict";` is handled by the compile-hook wrap and not generated by the compiler

* ESM output is cached even if output is identical to simplify loading logic
  *(since side-effect only modules are rare so shouldn't add overhead)*

* Runtime is only enabled for ESM
   *(I side step extending CJS by using `Runtime.prototype.runSetters`)*

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,9 +61,9 @@ module.watch(require("./module"), {
   // The keys of this object literal are the names of exported symbols.
   // The values are setter functions that take new values and update the
   // local variables.
-  default: value => { a = value; },
-  b: value => { b = value; },
-  c: value => { d = value; },
+  default(value) { a = value; },
+  b(value) { b = value; },
+  c(value) { d = value; },
 });
 ```
 
@@ -86,7 +86,7 @@ becomes
 ```js
 const utils = Object.create(null);
 module.watch(require("./utils"), {
-  "*": (value, name) => {
+  ["*"](value, name) {
     utils[name] = value;
   }
 });
@@ -113,7 +113,7 @@ becomes
 if (condition) {
   let b;
   module.watch(require("./c"), {
-    a: value => { b = value; }
+    a(value) { b = value; }
   });
   console.log(b);
 }
@@ -170,7 +170,9 @@ export { c as see }
 becomes
 
 ```js
-module.export({ see: () => c });
+module.export({
+  see: () => c
+});
 let c = 123;
 ```
 
@@ -203,9 +205,7 @@ around `module.export`:
 ```js
 module.exportDefault = function (value) {
   return this.export({
-    default: function () {
-      return value;
-    }
+    default: () => value
   }, true);
 };
 ```
@@ -287,8 +287,8 @@ export { a, b as c } from "./module";
 becomes
 ```js
 module.watch(require("./module"), {
-  a: value => { exports.a = value; },
-  b: value => { exports.c = value; },
+  a(value) { exports.a = value; },
+  b(value) { exports.c = value; },
 });
 ```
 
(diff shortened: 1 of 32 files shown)


jdalton:
Things to note:

* The `enforceStrictMode` is removed since `"use strict";` is handled by the compile-hook wrap and not generated.

* Runtime is only enabled for ESM (I side step extending CJS by using `Runtime.prototype.runSetters)

* The compiler `result` now includes a `sourceType` that specifies what time it compiled as

README.md
```diff
@@ -300,7 +300,7 @@ export * from "./module";
 becomes
 ```js
 module.watch(require("./module"), {
-  "*": (value, name) => {
+  ["*"](value, name) {
```
benjamn: This doesn't need square brackets, just quotes, I think.

lib/import-export-visitor.js
```diff
@@ -196,7 +183,9 @@ class ImportExportVisitor extends Visitor {
       decl.source.start
     ) + pad(
       this,
-      ',{"*":function(v,k){exports[k]=v}},' +
+      ",{" +
+      (this.generateShorthandMethodNames ? '["*"]' : '"*":function') +
```
benjamn: In other words, I believe you can remove the square brackets here.

README.md
```diff
@@ -61,9 +61,9 @@ module.watch(require("./module"), {
   // The keys of this object literal are the names of exported symbols.
   // The values are setter functions that take new values and update the
   // local variables.
-  default: value => { a = value; },
-  b: value => { b = value; },
-  c: value => { d = value; },
+  default(value) { a = value; },
+  b(value) { b = value; },
```
benjamn: Here's an example of why I avoided method shorthand before, even though it seems appealing at first. The assignment to `b` refers to the method name `b` here, not the outer variable `b`. Since this is the common case (`import { b } from ...`), I don't think we should use a syntax that works only for the uncommon case (`import { b as c } from ...`). Can we revert this commit?

node/compile-hook.js
```diff
@@ -4,64 +4,62 @@ const compiler = require("./caching-compiler.js");
...
+      : compiler.compile(code, { cacheFilename, cachePath, filePath, pkgInfo });
   }
 
+  const isModule = cacheValue.sourceType === "module";
   cache[cacheFilename] = cacheValue;
+  code = cacheValue.code;
+
+  if (isModule) {
+    code = 'module.run(function(){"use strict";' + code + "\n})";
+    Runtime.enable(mod);
```
benjamn: I think the runtime should be enabled for CJS as well, to support direct use of the runtime, for power-user cases.

lib/compiler.js
```diff
@@ -60,19 +58,14 @@ function compile(code, options) {
...
 
     importExportVisitor.finalizeHoisting();
+    result.identical = false;
+  }
 
-    if (! getOption(options, "repl")) {
-      result.code =
-        "module.run(function(){" +
-        magicString.toString() +
-        "\n})";
```
benjamn: The problem with removing this functionality from the compiler is that tools like Webpack or Meteor that use the compiler directly won't get the `module.run` wrapping, and may not be able to decide whether they should do it themselves, because they don't have access to the same information as the compiler.

jdalton:
Cool. I'll extract some of the non-controversial fixes/cleanup into their own PRs.

PR #179, opened 2017-05-25
Add minizlib guards for Node < 4.5.0.

jdalton:
This guards against the [current open issue](https://github.com/isaacs/minizlib/issues/2) of minizlib.

### DIFF
--- a/node/fs.js
+++ b/node/fs.js
@@ -1,9 +1,10 @@
 "use strict";
 
 const fs = require("fs");
+const minizlib = require("minizlib");
 const path = require("path");
 const utils = require("../lib/utils.js");
-const zlib = require("minizlib");
+const zlib = require("zlib");
 
 const FastObject = require("../lib/fast-object.js");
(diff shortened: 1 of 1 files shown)


node/fs.js
```diff
@@ -77,13 +88,22 @@ function gunzip(bufferOrString, options) {
...
-  const stream = new zlib.Gunzip(options);
-  if (options.encoding !== "utf8") {
-    return streamToBuffer(stream, bufferOrString);
+  if (useGunzipFastPath) {
+    try {
+      const stream = new minizlib.Gunzip(options);
+      if (options.encoding === "utf8") {
+        let result = "";
+        stream.on("data", chunk => result += chunk).end(bufferOrString);
+        return result;
```
benjamn: Is the `"data"` event always fired synchronously in this case?

jdalton: I suspect they'll fix the issue but until this having a fallback isn't a bad idea.

PR #202, opened 2018-02-05
Guide AST traversal based on a RegExp scan of possible indexes.

### LINKED ISSUE #113: Parse optimization idea.
From Slack:
> I think there’s another compiler optimization we could do
you can scan the source using a regular expression to find `import` and `export` tokens
then you can skip any AST nodes that don’t contain any of those tokens
and stop traversing once you know there won’t be any more tokens

benjamn:
When an AST traversal is looking for syntax that contains a known set of identifiers, such as `import` and `export`, we can ignore entire subtrees of the AST that do not contain any possible indexes of those identifiers, for a potentially massive savings in traversal time.

The list of possible indexes can be computed quickly using a regular expression, without actually parsing the code, since it's not the end of the world if there are false positives in the `possibleIndexes` array.

Closes #113, finally! :tada:

### DIFF
--- a/lib/visitor.js
+++ b/lib/visitor.js
@@ -151,17 +151,25 @@ class Visitor {
 
     for (let i = 0; i < keyCount; ++i) {
       const key = keys[i];
+
+      // Ignore .loc.{start,end} objects.
+      if (key === "loc") {
+        continue;
+      }
+
+      // Ignore "private" properties added by Babel.
+      if (key.charCodeAt(0) === codeOfUnderscore) {
+        continue;
+      }
+
       const value = node[key];
 
-      if (
-          // Ignore .loc.{start,end} objects.
-          key !== "loc" &&
-          // Ignore "private" properties added by Babel.
-          key.charCodeAt(0) !== codeOfUnderscore &&
-          // Ignore properties whose values aren't objects.
-          utils.isObject(value)) {
-        path.call(this.visitWithoutReset, key);
+      // Ignore properties whose values aren't objects.
+      if (! utils.isObject(value)) {
+        continue;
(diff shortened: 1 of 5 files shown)


jdalton:
This is so rad!

benjamn:
Should have trusted my instincts: a binary search at every level is a lot more complicated and expensive than a simple linear search that makes a little bit of progress at each level.

jdalton:
> Should have trusted my instincts: a binary search at every level is a lot more complicated and expensive than a simple linear search that makes a little bit of progress at each level.

Was wondering about that. When I saw the binary search algorthm being used my reaction was:

jdalton:
I'll try this out in `@std/esm` and see if I run into any problems.

lib/visitor.js
```diff
@@ -59,17 +151,25 @@ class Visitor {
 
     for (let i = 0; i < keyCount; ++i) {
       const key = keys[i];
+
+      // Ignore .loc.{start,end} objects.
+      if (key === "loc") {
+        continue;
+      }
```
jdalton: `loc` aren't there by default right, you have to opt-in to acorn producing them?

jdalton:
Tested in `@std/esm` without test fails! :tada:

lib/visitor.js
benjamn: That's right, but it's very important to skip them if they're there, since `loc` is a nested object (`loc.{start,end}.{line,column}`) attached to every node.

jdalton: Cool. Ya, just checking, I don't enable it and don't allow customization (to enable it).

PR #211, opened 2018-08-01
Create Entry objects before module evaluation in module.link.

benjamn:
With this commit, the Reify runtime no longer stores `Entry` objects on the `module.exports` object, but instead keeps them in a global map, keyed by absolute module identifiers.

By creating `Entry` objects before evaluating the corresponding modules, it's possible for modules involved in dependency cycles to access each other's `Entry` objects before they have finished evaluating, which allows sharing of hoisted function declarations, as required by the ECMA-262 specification.

Although it was previously possible to implement `module.link` in terms of `module.watch`, it is no longer possible to support `module.watch`, since it has no way of resolving the absolute module identifier of the child module, and thus cannot correctly populate the `Entry` map.

This commit essentially reverts PR #155, which abandoned an earlier global `Entry` map in favor of storing Entry objects on `module.exports`. At the time, I thought using `module.exports` would be simpler, in part because it made CommonJS "bridge" modules work:

```js
module.exports = require("./some/other/module");
```

Since then, however, we have accumulated a number of hacks to compensate for edge cases that I did not foresee, which is why this commit is clearly a simplification of the runtime implementation.

If you want the `module.exports` of one module to reflect the exports of another module (including the `default` export), the correct way to do that is now

```js
module.link("./some/other/module", { "*", "*+"});
```

### DIFF
--- a/lib/runtime/index.js
+++ b/lib/runtime/index.js
@@ -12,9 +12,12 @@ var Entry = require("./entry.js");
 // specific module objects, or for Module.prototype (where implemented),
 // to make the runtime available throughout the entire module system.
 exports.enable = function (mod) {
+  if (typeof mod.resolve !== "function") {
+    throw new Error("module.resolve not implemented");
+  }
+
   if (mod.link !== moduleLink) {
     mod.link = moduleLink;
-    mod.watch = moduleWatch;
     mod["export"] = moduleExport;
     mod.exportDefault = moduleExportDefault;
     mod.exportAs = moduleExportAs;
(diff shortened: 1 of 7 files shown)


lib/runtime/index.js
```diff
@@ -12,9 +12,12 @@ var Entry = require("./entry.js");
 // specific module objects, or for Module.prototype (where implemented),
 // to make the runtime available throughout the entire module system.
 exports.enable = function (mod) {
+  if (typeof mod.resolve !== "function") {
+    throw new Error("module.resolve not implemented");
+  }
```
benjamn: Not all module systems implement a `Module` constructor with a shared `Module.prototype`, and only some of those module systems implement `Module.prototype.resolve`. Since `module.link` now calls `module.resolve(id)` dynamically, a working implementation of `module.resolve` is now required. This is a significant requirement for sure, but hardly the deal-breaker that I once imagined it to be.

node/index.js
```diff
@@ -2,6 +2,11 @@
 
 const runtime = require("../lib/runtime");
 const setDefaults = require("../lib/options").setDefaults;
+const Module = module.constructor;
+
+Module.prototype.resolve = function (id) {
+  return Module._resolveFilename(id, this);
```
benjamn: Implementing `module.resolve` is easy in Node!

lib/runtime/index.js
```diff
@@ -29,35 +32,45 @@ exports.enable = function (mod) {
...
+  var exports = this.require(absChildId);
+
+  if (childEntry.module === null) {
+    childEntry.module = {
+      id: absChildId,
+      exports: exports
+    };
   }
+
+  childEntry.runSetters();
```
benjamn: Calling `childEntry.runSetters()` here is essential to report the incomplete exports of modules involved in import cycles, so that hoisted function declarations will be exposed as early as possible.

PR #229, opened 2019-06-18
Rename module identifier by wrapping program.body with IIFE.

benjamn:
Before this change, in order to obtain a reliable reference to the original `module` object, the Reify compiler would inject a variable declaration at the beginning of the generated code that bound some other (uniquely named) variable to the `module` object, and then that unique variable was used in all the other Reify-generated expressions, such as `module1.export(...)` and `module1.link(...)`.

The trouble with this strategy was that the injected variable declaration could still collide with other declarations in the module scope that involve the `module` identifer.

A better way to solve that problem is to introduce another scope by wrapping the module code in an immediately invoked function expression. The original `module` object can be passed as an argument to that function, and the corresponding parameter can be uniquely named:
```js
(function (module1) {
  module1.export({ name: () => name });
  module1.link("some-module", ...);
  // ...
}).call(this, module);
```

However, this wrapping should happen as late as possible, in some cases after multiple invocations of the Reify compiler, so the wrapping should not be the responsibility of the Reify compiler itself. Instead, the compiler should just use whatever `options.moduleAlias` is passed in, and not second-guess it or attempt to inject any variable bindings.

In this PR, the Babel plugin (`reify/plugins/babel`) becomes responsible for the IIFE wrapping shown above, while also gaining the ability to transform the AST twice: upon entering _and_ upon exiting the `Program` AST node. This double transformation is necessary to transform both `import` and `export` declarations found in the source code, and those that were injected later by other transforms, such as `@babel/plugin-transform-runtime`.

### DIFF
--- a/test/output/name/expected.js
+++ b/test/output/name/expected.js
@@ -1,4 +1,4 @@
-"use strict";var module1=module;module1.export({foo:()=>foo});module1.export({id:()=>id,name:()=>name},true);const path = require("path");
+"use strict";module.export({foo:()=>foo});module.export({id:()=>id,name:()=>name},true);const path = require("path");
 
 const id = module.id,
   name = path.basename(__filename);
--- a/lib/assignment-visitor.js
+++ b/lib/assignment-visitor.js
@@ -87,7 +87,7 @@ function wrap(visitor, path) {
         type: "MemberExpression",
         object: {
           type: "Identifier",
-          name: "module"
+          name: visitor.moduleAlias,
         },
         property: {
           type: "Identifier",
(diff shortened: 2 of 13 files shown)


test/output/name/expected.js
```diff
@@ -1,4 +1,4 @@
-"use strict";var module1=module;module1.export({foo:()=>foo});module1.export({id:()=>id,name:()=>name},true);const path = require("path");
+"use strict";module.export({foo:()=>foo});module.export({id:()=>id,name:()=>name},true);const path = require("path");
```
benjamn: As you can see from these updated tests, many of the `const module1 = module` renamings were unnecessary false positives before this change.

PR #242, opened 2020-12-01
Fix performance problem from throwing/catching lots of exceptions

jimrandomh:
`runGetter` is an inner-loop function that gets called every time anything is imported, and it showed up as a hotspot (~150ms) in my browser Javascript profiler results. Turns out, `getter` is sometimes null, which means that the small `try`/`catch` block here generates and consumes a null-reference exception. This is slow, much slower than handling the null case directly. Doing this reduces the time spent in this function to a much more reasonable ~30ms.

### DIFF
--- a/lib/runtime/entry.js
+++ b/lib/runtime/entry.js
@@ -393,6 +393,7 @@ function makeUniqueKey() {
 
 function runGetter(entry, name) {
   var getter = entry.getters[name];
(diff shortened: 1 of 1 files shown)


crapthings:
same here

https://github.com/benjamn/install/issues/60

we can even rewrite forEach to for loop

benjamn (approved):
Great idea—thanks @jimrandomh!

PR #245, opened 2021-02-22
exclude examples in production

crapthings:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "reify",
-  "version": "0.20.12",
+  "version": "0.20.13",
```
benjamn, line 3:
```suggestion
  "version": "0.20.12",
```
We'll do this on the `master` branch before publishing.

benjamn (approved):
Thanks @crapthings, this makes sense.

Issue #181, opened 2017-07-02
Seal namespace objects.


jdalton:
It looks like in the merge the load checks to seal the namespace object were dropped.

PR #246, opened 2021-03-29
Optimizations

zodern:
This PR tries to optimize:
- Importing a module that has already been imported a large number of times. Each time the module is imported, it loops through all previously added setters to check if any need to be called. When importing `@material-ui/icons`, most of the 3 minutes reify runs is spent on this
- Changing a single exported value in a module that is imported a large number of times. Svelte has a module that most svelte components import specific functions from, and the module also exports the component instance that is being initialized. Each time svelte creates a new component instance and changes that export, reify checks every setter to see if any need to be called, even though there usually are only a couple for the modified export. In one of my apps, this almost doubles the time svelte spends when switching pages.
- Re-exporting with `export ... from 'module'`. Many large npm packages that reify is slow at importing, such as `@material-ui/icons` and `mathjs` have a large number of these. When importing `@material-ui/icons`, ~20 seconds is spent running setters for parent modules, and most of the setters re-export values.

Changes:
- Check for changes once per name instead of once per setter. As far as I can tell, the only time some setters would have different previous values than others is if some of them are new and have never been ran. It now tracks the setters that have never been called separately so it can ensure they are still run.
- When using the `export ... from 'module'` syntax, the re-exported values are not available within the module. If the re-exported value changes, there is no reason to run the other getters since their value shouldn't be affected by it. When setters are created using the shorthand to re-export, Reify now records which exports the setter modifies. When those are the only setters ran for a parent, reify only runs the parent's getters for the modified exports instead of all of them. This optimization is currently not used for export all declarations (`export * from 'module'`).

### Performance improvement:

The times are from the second run to ensure the cache is up to date. I modified the package.json files for `@material-ui/icons` and `mathjs`  in `node_modules` to enable using reify with them.

#### Importing `@material-ui/icons` in Node 14

```js
import { Web, Add } from '@material-ui/icons/esm/index.js';
```
commonjs: 3 seconds
before: 3 minutes
after: 4 seconds. ~1/3 of the extra second compared to commonjs is from reify's Node compile hook
The changes also reduces memory usage by ~12mb.

```js
import * as icons from '@material-ui/icons/esm/index.js';
```
before: 3 minutes
after: 12 seconds

Tracking changes when running getters instead of when running setters would remove the difference between this and `import { Web, Add } from '@material-ui/icons/esm/index.js';`.

#### Importing `mathjs` in Node 14
```js
require('mathjs/lib/esm');
```

commonjs: 800ms
before:  1730ms
after: 1450ms

I haven't spent enough time looking into why `mathjs` is still slow to import to find the cause. It does have some export all declarations (`export * from 'module'`) which the re-export optimizations are disabled for.

### DIFF
--- a/lib/runtime/entry.js
+++ b/lib/runtime/entry.js
@@ -259,7 +259,7 @@ Ep.runSetters = function (names, runNsSetter) {
     setter(value, name);
   }
 
-  forEachSetter(this, names, undefined, runSetter);
+  forEachSetter(this, names, void 0, runSetter);
 
   if (this._uninitializedSetters.length > 0) {
     var initNames = [];
--- a/lib/runtime/entry.js
+++ b/lib/runtime/entry.js
@@ -26,6 +26,9 @@ function Entry(id) {
   // The normalized namespace object that importers receive when they use
   // `import * as namespace from "..."` syntax.
   this.namespace = utils.createNamespace();
+
+  this._lastValues = Object.create(null);
+  this._uninitializedSetters = [];
 }
 
 var Ep = utils.setPrototypeOf(Entry.prototype, null);
@@ -98,6 +101,12 @@ Ep.addSetters = function (parent, setters, key) {
         entry.setters[name] = Object.create(null);
       }
       entry.setters[name][key] = setter;
+
+      this._uninitializedSetters.push({
+        name: name,
+        key: key,
+        setter: setter
+      });
     }
   }
 
@@ -214,7 +223,8 @@ Ep.runSetters = function (names) {
   // module objects whose setters we might need to run.
   var parents;
 
-  forEachSetter(this, names, function (setter, name, value) {
+
+  function runSetter(setter, name, value) {
     if (parents === void 0) {
       parents = Object.create(null);
     }
@@ -223,7 +233,23 @@ Ep.runSetters = function (names) {
     // The param order for setters is `value` then `name` because the `name`
     // param is only used by namespace exports.
     setter(value, name);
-  });
+  }
+
+  forEachSetter(this, names, undefined, runSetter);
+
+  if (this._uninitializedSetters.length > 0) {
+    var initNames = [];
+    var initKeys = [];
+    this._uninitializedSetters.forEach(function (config) {
+      if (config.setter.initialized) {
+        return;
+      }
+      initNames.push(config.name);
+      initKeys.push(config.key);
+    });
+    this._uninitializedSetters.length = 0;
+    forEachSetter(this, initNames, initKeys, runSetter);
+  }
 
   if (! parents) {
     return;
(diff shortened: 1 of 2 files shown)


benjamn (commented):
@zodern Thanks for tackling this! It's refreshing to see a PR from someone who takes as much care as you do.

I will take a deeper look in the next few days, but I wanted to give you a heads up that I've cleaned up the repository a bit (enabled GitHub Actions and @dependabot, bumped dependencies, fixed a test, and closed/merged all the other PRs).

Would you mind rebasing? I think that will allow the tests to run on this PR, among other benefits.

benjamn:
@zodern Sorry for keeping you waiting here! I have a few commits I would like to push to this branch. Can you enable pushes to your branch? Alternatively, I could open another PR targeting your branch, but that's several more steps. Thanks!

zodern:
@benjamn the option should be enabled for you to push to this branch.

benjamn:
My mistake! Didn't catch that my local `zodern-master` branch was your `master` branch.

When you have a chance, I'd love to hear any thoughts you have on the commits I pushed. Still working a few other things, but I hope my adjustments make sense so far?

zodern:
I really like the changes you made. It does feel much safer.

I tried this branch in a Meteor app I'm working on that uses Svelte for some interactive visualizations, and I am trying to get it to run at 60 fps. Svelte updates its `current_component` export when a component is being updated. With the latest commit on this branch, reify adds ~12ms per frame. Switching to https://github.com/benjamn/reify/pull/246/commits/c4b2a93ce900327613d08270ef9e9f52eebd796e reduced it to ~6ms per frame. If I have some time this week or next I will look at this closer and see if we can reduce it.

StorytellerCZ:
Any update on this @zodern @benjamn

zodern:
I checked the performance again for `import { Web, Add } from '@material-ui/icons/esm/index.js';`, this time in a web browser (Edge 93.0.933.1). The times are a little slower than normal since I recorded a profile at the same time.

Time spent by `runSetters`:
- https://github.com/benjamn/reify/pull/246/commits/c4b2a93ce900327613d08270ef9e9f52eebd796e: ~1000ms
- https://github.com/benjamn/reify/pull/246/commits/f1ed7b3b3c5f9c394dce46454abe4b7f1a5e0251 - ~17750ms
- [snapshots with uninitialized](https://github.com/zodern/reify/tree/snapshots-with-uninitialized-tracking) - ~1000ms

These are two other branches with other experiments, but they seemed to have minimal if any effect when importing @material-ui (it still took ~1000ms):
- [optimization experiments](https://github.com/zodern/reify/tree/optimization-experiments)
- [track changes in runGetters](https://github.com/zodern/reify/tree/track-changes-in-run-getters)

### Svelte

I created a repro that has 102 components that cause there to be ~2,500 setters across 73 names for the `svelte/internal` module. For comparison, the largest svelte app I work on, which only uses svelte for a small part of the UI, has 1,548 setters for the `svelte/internal` module.

The times are how long `runSetters` takes during an update that causes the `current_component` export from `svelte/internal` to be modified 31 times.
- reify master: 36ms
- https://github.com/benjamn/reify/pull/246/commits/c4b2a93ce900327613d08270ef9e9f52eebd796e: 4.2ms
- https://github.com/benjamn/reify/pull/246/commits/f1ed7b3b3c5f9c394dce46454abe4b7f1a5e0251: 32.6ms
- [optimization experiments](https://github.com/zodern/reify/tree/optimization-experiments): 2.3ms
- [track changes in runGetters](https://github.com/zodern/reify/tree/track-changes-in-run-getters): 2.2ms

### Next steps

As is, the PR is much faster when importing large modules. If you think it is ready, we could merge it, and create separate PR's for any additional improvements.

We get a large performance improvement by reducing how often reify loops over each setter. I wasn't able to find a way to make it faster while keeping the looping. In the [snapshots-with-uninitialized-tracking](https://github.com/zodern/reify/commit/170142f252b8bc70bbdab1f591f02998b9a9c4a0) branch I re-implemented the uninitialized setters on top of your latest changes. I think this implementation should fix some of the problems with how I initially did it.

One commit in the [optimization-experiments](https://github.com/zodern/reify/commit/c00c1d4a6a89243893384f3e33a5e2d86e1be4fc) branch ([Cache getter and export names](https://github.com/zodern/reify/commit/c00c1d4a6a89243893384f3e33a5e2d86e1be4fc)) halves the remaining time spent in the Svelte reproduction. Caching export names would probably have to be done with #181, but they didn't have as much of an improvement as caching the getter names.

benjamn (approved):
I'm happy to merge this PR and release it, and continue iterating in follow-up PRs. Thanks again for pushing this forward @zodern!

benjamn:
@zodern @StorytellerCZ @filipenevola I've published these changes to npm as `reify@0.22.0`. You should now be able to depend on v0.22.0 in other packages, though I published this version with the `next` tag instead of `latest` for the time being (happy to update `latest` whenever you like). Cheers!

